Issue #9, opened 2021-12-14
Prompt for Notifications and Hash Selection should come before task selection


i128:
Right now, during the New Job Creation process, the user is prompted to select what tasks if any they'd like to include with the job, THEN they are prompted if they want to be notified upon a job creation / hash recovery. These steps should be reversed where the prompt for notifications/hash selection should come first before the task selection.

i128:
Closing

Issue #6, opened 2021-12-14
Top 10 Passwords and Passwords Count by Length Widths are not Dynamically adjusting with screen size.


i128:
When the page is rendered full screen, the widths of both bar charts render fine. But when the page is shown at half size, the graph extends beyond the card.

At full screen
![image](https://user-images.githubusercontent.com/2417131/146068416-b2056e1e-3ebc-495b-99b5-58b5d21fffb7.png)

At half screen:
![image](https://user-images.githubusercontent.com/2417131/146068476-12b3b925-409f-441b-8d29-4a0fd4f64771.png)

It would be nice to have the width of these graphs be dynamic to the width of the card they reside in.

i128:
Closing

PR #18, opened 2021-12-19
Code fixes for issue #6 and #9. As well as analytic page changes

StaticN0thing:
Fixed hashview issue #6 
Fixed hashview issue #9. 

Added config header to /analytics
Added chart download icons to /analytics
Added chart download icon hide toggle
Added Dynamic width to bar charts on /analytics
Added table join to prevent customers without hashes from showing up in the customer drop-down on /analytics
Added max height for dropdowns on /analytics so that large data sets will not overflow the page height

Moved adding tasks to after notification page and updated back and next links. (Please double check all the edge cases that may cause other redirects)

Added back notify_completed default to models.py (Not able to create jobs without it)
Added null check to jinja_hex_decode on __init__.py to prevent error if all hashes in file are already cracked

### DIFF
--- a/hashview/__init__.py
+++ b/hashview/__init__.py
@@ -21,7 +21,10 @@ scheduler = APScheduler()
 
 # Jinja2 Filter
 def jinja_hex_decode(text):
-    return bytes.fromhex(text).decode('latin-1')
+    if not text:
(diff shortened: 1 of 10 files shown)


i128:
Tested and everything looks good!

PR #37, opened 2022-03-25
Adding search export and display changes

StaticN0thing:
Analytics
    For All Customers
        Remove Total Accounts - Done
        Remove RunTime  - Done
        Both Bar graphs should render widths dynamically  - Was done already

Search
    Export Results  - Done
    
Notifications
    Include link with email/job notification - Was done already
    Convert # of seconds to days/hours/minutes, etc. - Was done already

Home
    make home page multiple jobs collapsible or only see your jobs - Made collapsible and added running status to the job header
    (not on list) User view will not have agents collapsed by default 
    
Jobs
	(not on the list) - Made the current users jobs show up at the top of the list always

CSS changes
	Removes double top border in information popups
	Making the buttons the same size so that when the page is resized they all line up
	Prevent long names from showing outside of popup modal
	Adding max height on dropdowns so they can't extend over the page view


i128 (approved):
Looks great!

PR #42, opened 2022-03-29
Last login date & authlib for password reset token handling

### LINKED ISSUE #15: Users List should show last login date
A new entry in the models.py for the users table should include the last date/time that they logged in. 
Then with that information, it should be displayed on the Manage->users list for administrators to review and cleanup for any stale users.

Yoshi325:
Happy to split these if desired.

The first commit is somewhat trivial, and hopefully provides the functionality desired to close #15 

The final commit is more complex, and has implications that should be reviewed thoroughly before merging (more info may be found in the body of the commit message).

### DIFF
--- a/hashview/models.py
+++ b/hashview/models.py
@@ -21,6 +21,7 @@ class Users(db.Model, UserMixin):
     admin = db.Column(db.Boolean, nullable=False, default=False)
     pushover_app_id = db.Column(db.String(50), nullable=True)
     pushover_user_key = db.Column(db.String(50), nullable=True)
+    last_login_utc = db.Column(db.DateTime, nullable=True, default=datetime.utcnow)
     wordlists = db.relationship('Wordlists', backref='tbd', lazy=True)
     rules = db.relationship('Rules', backref='owner', lazy=True)
     jobs = db.relationship('Jobs', backref='owner', lazy=True)
@@ -37,7 +38,7 @@ class Users(db.Model, UserMixin):
         try:
             user_id = s.loads(token)['user_id']
         except:
-            return None     
+            return None
         return Users.query.get(user_id)
 
 class Settings(db.Model):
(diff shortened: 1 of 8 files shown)


hashview/models.py
```diff
@@ -21,24 +24,78 @@ class Users(db.Model, UserMixin):
...
+    def _get_reset_token_salt(self) -> str:
+        """
+        Create salt data for password reset token signing. The return value will be hashed
+        together with the signing key. This ensures that changes to any of the fields included
+        in the salt invalidates any tokens produced with the old values.
+        """
+        return json.dumps([
+            self.first_name,
+            self.last_name,
+            self.password.hash.decode() if self.password is not None else '',
```
i128, line 43: So... i think this _should_ be 
`self.password if self.password is not None else '',`

Not entirely sure though

hashview/models.py
```diff
@@ -21,24 +24,78 @@ class Users(db.Model, UserMixin):
...
+            self.last_login_utc.to('UTC').isoformat() if self.last_login_utc else None
```
i128, line 44: This throws the following error:
```
AttributeError: 'datetime.datetime' object has no attribute 'to'
```

could we just do `self.last_login_utc if self.last_login_utc else None` ? do we need to convert the format if its all just going to be mashed into a salt?

i128: Ah, i see now why it needs to be converted :)

If i try what i suggested above i get the following error:
`TypeError: Object of type datetime is not JSON serializable`

so the datetime does need to be converted

hashview/models.py
Yoshi325: Yup. You're absolutely right. No need to use anything other than the raw string that the password is stored as.

hashview/models.py
Yoshi325: We should be able to drop the "to" since it is stored as UTC, and just covert to isoformat to then be serialized as a string.

Yoshi325:
In order to keep a clean git history, please let me know when you feel this has reached a stable point, and I can squish all these fixes into the proper commit they apply to before merging.

i128 (commented):
Everything looks great!
LMK if you still want to flatten this and if so i'll approve the merge.

PR #46, opened 2022-04-05
Pushover & Flask-Bcrypt replacement

### LINKED ISSUE #43: ImportError: cannot import name 'safe_str_cmp' from 'werkzeug.security'
[Flask-Bcrypt](https://pypi.org/project/Flask-Bcrypt/) appears to be hesitant to [cut a release](https://github.com/maxcountryman/flask-bcrypt/pull/66).

The published version on PyPI uses `safe_str_cmp` from `werkzeug` which was [deprecated](https://github.com/pallets/werkzeug/pull/2083) in [v2.0.0](https://werkzeug.palletsprojects.com/en/2.1.x/changes/#version-2-0-0)  (Released 2021-05-11) and then removed in [v2.1.0](https://werkzeug.palletsprojects.com/en/2.1.x/changes/#version-2-1-0) (Released 2022-03-28).

Because of this, I would suggest switching to a maintained fork of it: [Bcrypt-Flask](https://pypi.org/project/Bcrypt-Flask/). It appears to be a drop in replacement.

Another alternative would be to inline the bcrypt functionality required.

Yoshi325:
The latest published version of the python-pushover package requires `use_2to3` which is no longer provided by [setuptools](https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v5800). In order to resolve this, I walked through the package, as well as the Pushover API, and came to the conclusion that it would be better to use the API directly. (it requires only a handful of lines, instead of an additional dependency).

More info about the Flask-Bcrypt to Bcrypt-Flask switch can be found in the related Issue.

Closes #43

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Notable changes will be documented here
 - Moved SMTP settings to config file
 - Hashview Agent is now packaged with hashview (server) under install directory. Can be downloaded from agents menu as admin
 - Changed from itsdangerous to authlib for password reset token generation (please make sure to update your environments to include authlib).
+- Changed from the python-pushover package, to a call directly to the Pushover API.
 ### Removed
 - Removed hashview agent from local processing. If you want to run hashview AND crack hashes on the same box run the hashview-agent in a seperate screen/tmux session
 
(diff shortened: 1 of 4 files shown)


i128 (approved):
Tested out and everything looks great! Thanks!

PR #70, opened 2022-06-27
add hashtype

Hafiziahmad:


i128:
I appreciate the contribution! I need to add some sanity checks on these hashes so that users cant/dont submit a wrong hash/ hash_type combinations. Once i get that sorted out i'll merge the request!

Hafiziahmad:
Thank you for give opportunity

ovcrash:
> I appreciate the contribution! I need to add some sanity checks on these hashes so that users cant/dont submit a wrong hash/ hash_type combinations. Once i get that sorted out i'll merge the request!

I don't think that there is a good way to do this. Like they said on the hashcat FAQ.

_Tools which claim to be able to identify hashes simply use regular expressions to match the hash against common patterns. This method is extremely unreliable and often yields incorrect results. It is best to avoid using such tools.

A much better way to identify the hash algorithm would be to understand the origin of the hashes (e.g. operating system, COTS application, web application, etc.) and make an educated guess at what the hash algorithm might be. Or better yet, use the source!_

source: https://hashcat.net/wiki/doku.php?id=frequently_asked_questions#how_can_i_identify_the_hash_type

i128:
I agree, and thats essentially what hashview does. We've found that its better to have SOME sanity checks on hash types vs none even if they're not guaranteed.

PR #72, opened 2022-07-25
modify run_hashcat to use subprocess.run()

Infosec-McClane:
This PR contains the following modifications:

install/hashview-agent/hashview-agent.py (line 315): switch to using subprocess.run() for executing hashcat. This accepts an array to parameterize the command string.

install/hashview-agent/hashview-agent.py (line 314): Use python function open() to create the log file output, used by line 315.

utils/utils.py (line 218 - 233): modified the command strings to be an array that can be ingested by subprocess.run().

hashview/jobs/routes.py (line 423): cast the cmd to a string so it can be saved to MySQL for distributed computing.


i128:
Looks great! Great work!

PR #97, opened 2022-10-24
Implemented exception handling to agent

osean-man:
Running into an error where the return was None, an exception would crash the agent. Implemented a fix with exception handling.

### DIFF
--- a/install/hashview-agent/agent/api/api.py
+++ b/install/hashview-agent/agent/api/api.py
@@ -118,15 +118,18 @@ def updateJobTask(job_task_id, task_status):
         'job_task_id': job_task_id
     }
 
-    response = http.post('/v1/jobtask/status', json.loads(json.dumps(message)))
-    decoded_response = json.loads(response)
-    if decoded_response['type'] == 'message' and decoded_response['status'] == 200:
-        return decoded_response
-    elif decoded_response['type'] == 'message' and decoded_response['status'] == 426:
-        print('Our agent version is older than the servers. You need to upgrade your agent before continuing.')
-        exit()
-    else:
-        print('We got an unexpected response type or status code.')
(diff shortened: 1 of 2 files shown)


i128:
Interesting. I appreciate the PR. Do you recall what situation this happened in? Like was it in-between agent call backs and the user killed/stopped a task/job?

osean-man:
I loaded the following hash: c05e35377b5a31f428ccda9724a9dfbd0c5d71dccac691228d803c78e2e8da29 (as an upload).

It was just a test and I got this from the NahamStore (TryHackMe room). It's a raw sha-256 (HC mode 1400). The agent crashed on the TypeError as the json.loads received a NoneType. I suppose rather than a try block it could just default return an empty array instead. This was just my quick fix while trying out Hashview. 

If you need to reproduce, I did the following:

- Fresh install of hashview (note: I did have an issue with SQLAlchemy/collation that I had to solve with this https://tecadmin.net/resolved-unknown-collation-utf8mb4_0900_ai_ci/)
- Created an agent and approved
- Created a job for the agent
- Agent crashed on the line numbers this PR is for with a TypeError (NoneType not supported...).

i128:
Awesome, i appreciate the feedback, I will attempt to reproduce. 
Regarding the block try vs returning an empty array, i almost prefer the error so that i can handle those use cases. Because in theory that value should never be none. But i agree this shouldnt be happening in the first place.

osean-man:
Agreed. No problem. I figured the program should continue rather than die so I used https://docs.python.org/3/library/contextlib.html#contextlib.suppress to handle the TypeError specifically. 

Great work on the project, I'm looking forward to using it.

i128:
Merged, sorry for the delay

PR #104, opened 2023-01-17
Add docker infrastructure

Yoshi325:
This is a pretty heavy one. Had to shift a bunch around to make it work. Specifically the command line setup needed to be split to support a web ui setup if user input is required. And the other half of it needed automatically (but idempotently) run before first request.

With a proper `hashview/config.conf`: `docker compose up` will bring up a container for mysql, as well as one for the application. A `before_first_request` hook sets up most of the default items. And a `before_request` hook handles changing the Admin user that is created by default, and populating the Settings in the database.

While working, I also addressed two comparisons with 0 that Flask was flagging. And unified logging a bit so that a common formatting is used, as well as fixing a bug where the flask migrate upgrade was squelching further log messages.


hashview/__init__.py
```diff
@@ -8,6 +18,121 @@ def get_application_version() -> str:
...
+    from flask import current_app
+    logger = current_app.logger
+    logger.info('Setting up defaults on before first request.')
+
+    from hashview.models import db
+
+    try:
+        logger.info('Upgrading Database if needed Progressing.')
+        import alembic.command
+        migrate_ext = config = current_app.extensions['migrate']
```
Yoshi325, line 69: I need to remove this double-assignment.

hashview/__init__.py
```diff
@@ -8,6 +18,121 @@ def get_application_version() -> str:
...
+    if parsed_url.path.startswith(static_path):
+        # allow static files through
+        return
+
+    from hashview.models import db
+
+    from hashview.setup import admin_pass_needs_changed
+    from hashview.users.routes import bcrypt
+    if not admin_pass_needs_changed(db, bcrypt):
+        logger.info('Admin password needs changed.')
```
Yoshi325, line 38: I need to fix this log message (it should convey the inverse).

Yoshi325:
I need to add a section the the README. Specifically it should call out that this first pass of docker is intended to be a platform to build future production deployment of hashview via docker, but for now it is in more of a "developer ready" state. Also, it should point out that a `config.conf` needs created in order to be able to run `docker-compose up` and that to be secure, the `MYSQL_PASSWORD` should be changed in `docker-compose.yml`.

Yoshi325:
I need to fit the typo in my latest commit (ducker => docker).

bandrel:
Getting errors when trying to docker compose up. 
image
I had to change the port to 5001:5001 because of a conflict other than that the example config was used.

NotoriousRebel:
Any updates on this @Yoshi325 ?

i128:
This is the next update to go in. We wanted to land v0.8.1 first.

Yoshi325:
Rebased and updated on top of the latest.

hashview/__init__.py
```diff
@@ -1,9 +1,140 @@
...
+        from hashview.users.routes import bcrypt
+        from hashview.setup import add_admin_user
+        from hashview.setup import admin_user_needs_added
+        if admin_user_needs_added(db):
+            logger.info('Adding Admin User.')
+            add_admin_user(db, bcrypt)
+    except:
+        logger.exception('Adding Admin User failed.')
+
+    try:
```
i128, line 99: We should swap the order in which the default wordlists are added. Starting with the dynamic wordlist followed by the static wordlist. 

This way, the dynamic wordlist, in the db ends up with id=1.

hashview/setup/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+
+def default_tasks_need_added(db :SQLAlchemy) -> bool:
+    return (0 == db.session.query(Tasks).count())
+
+
+def add_default_tasks(db :SQLAlchemy):
+    task = Tasks(
+        name          = 'Rockyou Wordlist',
+        owner_id      = '1',
+        wl_id         = '1',
```
i128, line 28: This is correct for now, in so much as the wl_id=1 is rock you, but if we swap it so that the dynamic wordlist is loaded first, then these should be updated wl_id=2

hashview/setup/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        rule_id       = None,
+        hc_attackmode = 'dictionary',
+    )
+    db.session.add(task)
+
+    task = Tasks(
+        name          = 'Rockyou Wordlist + Best64 Rules',
+        owner_id      = '1',
+        wl_id         = '1',
```
i128, line 37: Same here. update to wl_id=2 if we make dynamic wordlist wl_id=1

hashview/setup/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        wl_id         = '2',
+        rule_id       = None,
+        hc_attackmode = 'dictionary',
+    )
+    db.session.add(task)
+
+    task = Tasks(
+        name          = 'Rockyou Wordlist + Best64 Rules',
+        owner_id      = '1',
+        wl_id         = '3',
```
i128, line 37: This should be wl_id 2, not 3.

Issue #114, opened 2023-08-19
Syntax error in hashview.utils.utils:110 v0.8.1


Yoshi325:
https://github.com/hashview/hashview/blob/998eb824e5487ec21a32fa952cd8987194c1732e/hashview/utils/utils.py#L110C1-L110C60

PR #115, opened 2023-08-19
:bug: Fix Issue #114

Yoshi325:
a bunch of trailing whitespace was trimmed as well


i128:
Dangit, good catch on all of these. thanks!

PR #120, opened 2023-09-22
Update routes.py

vsamiamv:
Updated the dictionary section of the tasks so that you can create a task with no rule (None) as Default then change the edit code so that the interface shows the options that are in the database and lets you select what is in the DB as apposed to just showing all values in the DB.

### DIFF
--- a/hashview/tasks/routes.py
+++ b/hashview/tasks/routes.py
@@ -31,13 +31,14 @@ def tasks_add():
 
     for wordlist in wordlists:
         tasksForm.wl_id.choices += [(wordlist.id, wordlist.name)]
-
+    
+    tasksForm.rule_id.choices = [('None', 'None')]
     for rule in rules:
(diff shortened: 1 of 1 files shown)


i128:
This is great, and honestly something that should have been in place from the get go. In fact i could have sworn it was. 
https://github.com/hashview/hashview/blob/main/hashview/utils/utils.py#L241

This shows that, when building the hashcat command that gets sent down to the agents, that we handle conditions where there are no rule's assigned to a dictionary based attack. I wonder if we had it once, and it got removed? Either way I appreciate the submission! TYTY

vsamiamv:
Rockin.. yeah there is a simple bug in the "masking" update also.. 

I think on line 62  hc_mask=tasksForm.mask.data there is a comma at the end.. also... line 140 has hc_mask instad of task.hc_mask so if you update it dosen't go to the DB

vsamiamv:
imma keep adding junk .. i need it to post to slack and discord..

PR #125, opened 2024-04-26
Adds Docker support

puzzlepeaches:
A continuation of https://github.com/hashview/hashview/pull/104 that fully works. 

This MR adds docker and docker-compose support to hashview. To test, do the following:

1. Install docker and docker compose.
2. Copy `.env.example` from the docker directory and update with your wanted environment variables in the root of the project.
3. Run `docker compose build --no-cache`
4. Run `docker compose up` / `docker compose up -d`
5. Navigate to the browser and ensure you can log in with the admin credentials specified in your `.env` file.

Notes:

- The MySQL container is tied to a volume. If the build fails for whatever reason, you must delete the containers AND volume to rebuild successfully. `--no-cache` on your build will not work.
The updates to README.md are minimal. We may need to do more to support its use in the real world.
- This MR does not support the agent. It currently must be installed bare metal. 
- Containers are built using `linux/amd64` as the platform. This can be changed by removing the platform lines from the `docker-compose.yaml` and `Dockerfile` respectively.


i128:
When you're ready for this to get merged, please submit this PR to the v0.8.2-dev branch and not main.

PR #126, opened 2024-04-27
Upload wordlists/rulesets from disk

puzzlepeaches:
This PR adds support for uploading wordlists/rulesets from disk instead of a web browser. The functionality here is identical to the Ruby version. 

Changes Made:

Directories were added to the following locations and contain `.gitkeep` files:

```
hashview/control/rules/auto_upload
hashview/control/wordlists/auto_upload
```

A job with APScehduler was added to `hashview.py` for both rulesets and wordlist that checks for new files every minute. When a new file is detected in the auto_upload folders:

1. A checksum for the file is generated and compared to existing rulesets/wordlists in the DB
2. If the wordlist/ruleset is unique, the file is imported into the database in the background

Notes:
- This PR contains the code from https://github.com/hashview/hashview/pull/125. I think you should merge that first before attempting to merge this code. 
- This feature has yet to be tested with very large files. Can't with my current setup.
- The one-minute interval check for new word lists might be too frequent. I cannot tell how system-intensive the file hashing is on a large wordlist. If it eats a ton of resources, the user may experience issues.
- Changes were also made to the `docker-compose.yaml` file to support this change for a docker-based environment. They are commented out with a note.


i128:
When you're ready for this to get merged, please submit this PR to the v0.8.2-dev branch and not main.

PR #137, opened 2024-11-19
Optimize the jobs list page a little

Yoshi325:
so that more of the filtering is done in python beforing being passed to jinja to render the template


Yoshi325:
Added a fix for a problem discovered by @bandrel while testing.

i128:
Only closing since i think this has already been addressed.

PR #154, opened 2026-02-03
init commit of pytest framework

bandrel:
Initial commit of pytest framework including playwright. Intent is to use for future development and regression testing. The tests also include some basic security checks such as XSS sanitization checks.


Yoshi325:
Looks good.

I think it would be very beneficial to consider adding:
- a section to the readme about how to run the tests (I like the documentation in TESTING.md, but as a dev I normally look at the README first).
- possibly a shell script like `run-tests.sh` to add discoverability (it could facilitate the setup detailed in TESTING.md, since there are a lot of steps that could go wrong).
- a GitHub action to run the tests

Or we can open Issues for any/all of those to be done in another pass.

What do you think?

bandrel:
TESTING.md is referenced in [README.md](https://github.com/hashview/hashview/blob/tests/pytest-playwright/README.md#testing) maybe i wasnt verbose enough 

[run_e2e_compose.sh](https://github.com/hashview/hashview/blob/tests/pytest-playwright/tests/run_e2e_compose.sh) contains the shell script to do what you're asking. it is just in the tests folder and not the top-level. we can move it if you think it should be moved. 

I didnt add the github action yet because of the existing github actions that are currently failing. I figured we would add it after we merge the dev branch into main

Yoshi325 (approved):
Good first stab at testing. I'm excited that we are beginning to test.

PR #189, opened 2026-06-10
Add server-side conversion for WPA captures and NTDS.dit files

Yoshi325:
Allows users to upload raw WPA capture files (pcap/pcapng from pwnagotchi or
hcxdumptool) and NTDS.dit + SYSTEM hive pairs directly. Hashview converts them
server-side and imports the resulting hashes asynchronously, so the upload
returns immediately without a long wait.

## How it works

A new `HashfileConversions` table tracks conversion state for each upload. An
APScheduler job polls every 30 seconds, runs the appropriate conversion tool,
and feeds the output into the existing hash import pipeline:

- **WPA captures** → `hcxpcapngtool` → hashcat mode 22000 (WPA-PBKDF2-PMKID+EAPOL)
- **NTDS.dit + SYSTEM hive** → `impacket-secretsdump LOCAL` → hashcat mode 1000 (NTLM)

Source files are removed from disk after a successful conversion. Failed
conversions are marked with an error message, shown as a status badge on the
hashfiles list page.

## Dependencies

- `hcxtools` (`apt install hcxtools`) — required for WPA capture conversion
- `impacket` (`pip install impacket`) — required for NTDS.dit conversion

Both are optional at runtime; uploading the relevant file type on a server
without the tool installed produces a clear error message.

## Changes

- New model `HashfileConversions` with an Alembic migration (chained onto the
  current `v0.8.3-dev` head).
- New module `hashview/utils/convert.py` (`convert_pcap`, `convert_ntds`,
  `ConversionError`).
- New scheduler job `process_pending_conversions` (30-second interval).
- Upload form: two new `file_type` options (`wpa_pcap`, `ntds`); NTDS uploads
  include a second file field for the SYSTEM hive.
- Hashfiles list: status badges for pending / converting / failed conversions.
- 28 unit tests covering the failure modes, file cleanup, cascade deletion, and
  the full upload → scheduler → import path.

## REST API

Three endpoints provide programmatic access to the same capability, documented
in the OpenAPI spec (`hashview/api_docs/openapi.yaml`) and covered by 15 unit
tests (auth rejection, missing fields, invalid customer IDs, and the happy path
for each):

- `POST /v1/hashfiles/upload/wpa_pcap/<customer_id>/<hashfile_name>` — raw
  binary body (pcap/pcapng); returns `hashfile_id` and `conversion_id`.
- `POST /v1/hashfiles/upload/ntds/<customer_id>/<hashfile_name>` —
  multipart/form-data with `ntds_file` and `system_file`; returns `hashfile_id`
  and `conversion_id`.
- `GET /v1/hashfiles/<hashfile_id>/conversion` — poll conversion state; returns
  `conversion_status`, `source_type`, and `conversion_error`.

All three are user-authenticated (the existing `uuid` cookie / API key). The
intended flow is upload → poll → done.

## Also included (outside the feature)

- **Resolved an Alembic dual-head** that appears when this branch merges into
  `v0.8.3-dev`: the base added migration `d4e8b1f3a297` (task-loopback) off the
  same parent as the conversion migration, producing two heads — which silently
  aborted the startup `upgrade`, left an empty schema, and failed the e2e run.
  The conversion migration is re-chained onto `d4e8b1f3a297` so history stays
  linear.
- **Fixed the `DATA_RETENTION` scheduler job**, which had the same
  `current_app`-proxy bug as the new conversion job (jobs ran outside an app
  context and silently failed). Pre-existing; fixed here because the new
  30-second job exposed the pattern.
- **Hardened the DB healthcheck** in `docker-compose.yml` to gate on a real TCP
  query as the app user instead of the socket-only `mysqladmin ping`, so the app
  doesn't start before MySQL is truly ready. Affects `docker compose up` as well
  as CI.
- **Repaired stale e2e tests** for the current UI: strict-mode "New Job"
  locators, and `test_tasks_edit` rewritten for the modal-based edit flow;
  `test_agent_sim` now launches via `sys.executable` for host portability.
  `tests/run_e2e_compose.sh` also dumps container logs on a readiness timeout so
  boot failures are diagnosable.
- **Added `scripts/preflight.sh`** to run all CI gates locally before pushing,
  and pinned `pylint` in `requirements-dev.txt`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,4 @@ hashview/control/hashes/*
 hashview/control/logs/*
 hashview/control/wordlists_import/*
 hashview/ssl/*
+.scratch/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,9 +26,18 @@ services:
     volumes:
       - db:/var/lib/mysql
     healthcheck:
-      test: ["CMD", "mysqladmin" ,"ping", "-h", "localhost"]
-      timeout: 20s
-      retries: 10
+      # Gate on the REAL server: force TCP (-h 127.0.0.1) and run a query as the
+      # app's user against its database, so the check only passes after init is
+      # fully done. The previous `mysqladmin ping -h localhost` used the unix
+      # socket, which answers during MySQL's temporary init server — letting the
+      # app start and run migrations before the real server accepts TCP. Those
+      # migrations fail (connection refused) and get swallowed, leaving an empty
+      # schema and "Table 'hashview.users' doesn't exist" on every request.
+      test: ["CMD-SHELL", "mysql -h 127.0.0.1 -u hashview -phashview -e 'SELECT 1' hashview || exit 1"]
+      interval: 5s
+      timeout: 5s
+      retries: 20
+      start_period: 30s
 
 volumes:
   db:
--- a/hashview/__init__.py
+++ b/hashview/__init__.py
@@ -69,15 +69,30 @@ def setup_defaults_if_needed():
         logger.exception('Upgrading Database failed.')
 
     try:
-        from hashview.scheduler import data_retention_cleanup, scheduler
+        from hashview.scheduler import (
+            data_retention_cleanup,
+            process_pending_conversions,
+            scheduler,
+        )
         logger.info('Clearing Scheduled Jobs.')
         scheduler.remove_all_jobs()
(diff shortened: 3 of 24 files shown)


Yoshi325:
Working on adding API elements for this, and fixing the failing checks.

Yoshi325:
We are going to pass on this for now.

Issue #217, opened 2026-06-15
API: /v1/jobs/start can never start an API-created job (inverted status guard)


bandrel:
## Summary
`v1_api_post_start_job` (`hashview/api/routes.py:808-851`) bails out if `job.status != 'Queued'`, but `/v1/jobs/add` creates jobs with `status='Ready'` (`routes.py:735`). The web UI `jobs_start` (`hashview/jobs/routes.py:858`) has no such precondition.

**Result:** a job created through the API is stuck — it is `Ready`, and start rejects anything not already `Queued`, so it can never be started via the API.

## Location
- `hashview/api/routes.py:808-851`

## Suggested fix
The guard is effectively inverted. It should reject jobs already `Running`/`Queued` and allow `Ready` (and similar) jobs to be queued, matching the web UI behavior.

PR #247, opened 2026-06-19
fixed issue #217

i128:


bandrel:
Fixed the failing unit test in 8b881fa.

**Root cause:** the test `test_jobs_start_returns_400_when_job_not_queued` was written against the pre-fix (inverted) guard. With the issue #217 fix in place, a `Ready` job is now correctly allowed to start, so it fell through into `build_hashcat_command` with the test's bogus `task_id=1` and raised `AttributeError: 'NoneType' object has no attribute 'rule_id'`.

**Fix:** updated the test to assert the 400 path that still exists under the new logic — a job that is already `Queued`. Renamed to `test_jobs_start_returns_400_when_job_already_queued`. No production code changed; the fix for #217 is correct.

PR #252, opened 2026-06-19
fix: correct search export passing wrong variable and undefined buffer (#211)

### LINKED ISSUE #211: Search results Export is broken (passes None instead of results)
## Bug
`hashview/searches/routes.py:69` calls `export_results()` with the local `results`, which is always `None` on the POST/export path; the actual data is in `hashfile_results`:

```python
export_results(results)   # results is None here -> TypeError inside get_rows()
```

Clicking **Export** on a search raises `TypeError: 'NoneType' object is not iterable`. The export feature is effectively non-functional.

## Fix
Pass the populated variable:
```python
export_results(hashfile_results)
```
(verify the intended variable name against the surrounding code)

## Repro / regression test
Strict `xfail` in `tests/unit/test_rules_searches_notifications_branches.py` on branch `test/route-branch-coverage`.

MohammadYusif:
## Summary

- The search-results Export feature was completely non-functional and raised an exception before returning a file.
- Two compounding name errors on the export path in `hashview/searches/routes.py`:
  - `searches_list()` passed the local `results` (undefined/None on the POST and `hash_id` paths) to `export_results()` instead of the populated `hashfile_results`.
  - `export_results()` and `get_rows()` referenced an undefined `strIO` while the actual buffer variable is `str_io`, raising `NameError`.

## Changes

- `hashview/searches/routes.py`: pass `hashfile_results` into `export_results()`; correct the `strIO` references to `str_io` in `export_results()` and `get_rows()`; drop the now-dead `results = None` branch.

This also makes the strict `xfail` added for #211 in #214 pass.

Fixes #211

### DIFF
--- a/hashview/searches/routes.py
+++ b/hashview/searches/routes.py
@@ -54,10 +54,9 @@ def searches_list():
             flash('No results found.', 'warning')
     else:
         customers = None
-        results = None
 
     if hashfile_results and "export" in request.form: #Export Results
-        return export_results(customers, results, hashfiles, searchForm.export_type.data)
+        return export_results(customers, hashfile_results, hashfiles, searchForm.export_type.data)
 
     return render_template('search.html.j2', title='Search', searchForm=searchForm, customers=customers, hash_results=hash_results, hashfile_results=hashfile_results, hashfiles=hashfiles, redacted_data=redacted_data)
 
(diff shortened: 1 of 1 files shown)


i128:
Looks good. thank you

PR #278, opened 2026-06-22
Re introduce chunks

i128:


hashview/agents/routes.py
```diff
@@ -11,15 +11,67 @@
...
+def _fmt_speed(hps):
+    """Raw hashes/sec -> human string, e.g. 28460000000 -> '28.46 GH/s'."""
+    value = float(hps or 0)
+    idx = 0
+    while value >= 1000 and idx < len(_SPEED_UNITS) - 1:
+        value /= 1000.0
+        idx += 1
+    if idx == 0:
+        return '%d %s' % (int(value), _SPEED_UNITS[idx])
+    return '%.2f %s' % (value, _SPEED_UNITS[idx])
```
Yoshi325:
```python
from typing import Final

_SPEED_UNITS: Final = ('H/s', 'kH/s', 'MH/s', 'GH/s', 'TH/s', 'PH/s')


def _unit_power(hps: float) -> int:
    """Index of the largest unit hps has reached (0 = raw H/s, including zero)."""
    reached = [power for power in range(len(_SPEED_UNITS)) if hps >= (1000 ** power)]
    return max(reached, default=0)


def _fmt_speed(hps: int | float | None) -> str:
    """Raw hashes/sec -> human string, e.g. 28460000000 -> '28.46 GH/s'."""
    hps = float(hps or 0)
    power = _unit_power(hps)
    value = hps / (1000 ** power)
    unit = _SPEED_UNITS[power]
    if power == 0:
        return f'{int(value)} {unit}'
    else:
        return f'{value:.2f} {unit}'
```

might provide a little better clarity on what is happening here

however, it looks like this duplicates functionality from utils.fmt_hps so we might consider using that instead

hashview/agents/routes.py
```diff
@@ -11,15 +11,67 @@
...
+            .order_by(AgentBenchmarks.hash_type)
+            .all())
+    for row in rows:
+        out.setdefault(row.agent_id, []).append({
+            'hash_type': row.hash_type,
+            'name': _MODE_NAMES.get(row.hash_type, ''),
+            'speed': _fmt_speed(row.speed),
+            'updated_at': row.updated_at,
+        })
+    return out
```
Yoshi325: We might consider refining `_agent_benchmarks`.

It currently keeps two parallel structures for "which agents" (`ids` list + `out` dict), and the `if not ids` guard reads as "return an empty thing" when its real job is to skip the empty-`IN()` query. Collapsing to one structure makes the intent obvious and removes the dead `setdefault` (every queried `row.agent_id` is already a seeded key, so its default branch can never run):

```python
def _agent_benchmarks(agents_):
    """
        agent.id -> list of {hash_type, name, speed, updated_at} for the modal.
        
        Every agent is present; agents with no benchmarks map to [].
    """
    if (not agents_):
        # no agents -> skip the empty-IN() query
        return {}
    else:
        benchmarks_by_agent_id = {
            a.id: [] 
            for a in agents_
        }
        benchmarks = (
            AgentBenchmarks.query
                .filter(
                    AgentBenchmarks.agent_id.in_(
                        benchmarks_by_agent_id.keys()
                    )
                )
                .order_by(
                    AgentBenchmarks.hash_type
                )
                .all()
        )
        for record in benchmarks:
            record_as_dict = {
                'hash_type':  row.hash_type,
                'name':       _MODE_NAMES.get(row.hash_type, ''),
                'speed':      _fmt_speed(row.speed, places=2),
                'updated_at': row.updated_at,
            }
            benchmarks_by_agent_id[row.agent_id].append(record_as_dict)
        return benchmarks_by_agent_id
```

hashview/agents/routes.py
```diff
@@ -144,20 +222,27 @@ def agents_delete(agent_id):
...
+    # Remove every reference to this agent BEFORE deleting it. Two foreign keys
+    # point at agents.id — job_tasks.agent_id and agent_benchmarks.agent_id — and
+    # either one blocks the delete (the commit rolls back, so the agent silently
+    # reappears). Re-queue any chunk it was actively running so the work isn't
+    # orphaned, unlink the rest, and drop its stored per-hashtype benchmarks.
+    for jt in JobTasks.query.filter_by(agent_id=agent_id).all():
+        if jt.status == 'Running':
+            jt.status = 'Queued'
+        jt.agent_id = None
+    AgentBenchmarks.query.filter_by(agent_id=agent_id).delete(synchronize_session=False)
```
Yoshi325, line 227: I think some drift might have happened, this comment does not align with the code after it.

hashview/main/routes.py
```diff
@@ -111,26 +117,188 @@ def _agents_ctx():
         'agents': agents,
         'recovered_list': recovered_list,
         'time_estimated_list': time_estimated_list,
+        'telemetry': agent_telemetry(agents),
     }
 
 
+# Crack-rate helpers. _hps/_fmt mirror the (nested, non-importable) sidebar helpers
+# in hashview/__init__.py; kept module-level here so the per-task chunk aggregation
+# can sum agent speeds without a circular import on the package root.
```
Yoshi325: This and lines 131+132 seem to conflict. I think these three lines should be deleted.

hashview/utils/utils.py
```diff
@@ -650,8 +652,146 @@ def update_dynamic_wordlist(wordlist_id, job_id=None):
...
+
+
+def parse_hps(s):
+    """Parse a speed string to raw hashes/sec, tolerantly.
+
+    Accepts '284.6 GH/s', lowercase 'gh/s', thousands separators ('1,024 MH/s'),
+    trailing text ('284.6 GH/s (12ms)'), units k/M/G/T/P/E, a bare number (assumed
+    H/s), or an int/float. Returns 0.0 when no number is present. Single source for
+    the dashboard task-rate sum, the sidebar aggregate, and agent utilization.
+    """
```
Yoshi325, line 693: It looks like this isn't the single source, there are duplicate implementations.

hashview/models.py
```diff
@@ -225,6 +244,31 @@ class Agents(db.Model):
...
+    # short label (e.g. 'RTX 4090'); gpu_temps is a comma-separated list of the
+    # per-card temperatures in °C (e.g. '71,70,72').
+    gpu_model = db.Column(db.String(128))
+    gpu_temps = db.Column(db.String(128))
+
+class AgentBenchmarks(db.Model):
+    """Per-(agent, hash_type) hashcat benchmark used to size task chunks.
+
+    `speed` is raw hashes/sec summed across the agent's devices, parsed from
+    `hashcat -b -m <mode> --machine-readable`. The chunk planner sizes chunks
```
Yoshi325: Don't we use `--json` rather than machine readable for the benchmarks? If so, this docstring is not correct.

hashview/api/routes.py
```diff
@@ -273,8 +338,15 @@ def v1_api_set_agent_heartbeat():
                     }
                     return jsonify(message)
 
-                if settings.max_runtime_tasks > 0 and job_task.started_at is not None and job_task.started_at + timedelta(hours=settings.max_runtime_tasks) < datetime.now():
-                    update_job_task_status(job_task.id, 'Canceled')
+                # Enforce max_runtime_tasks on the PARENT task, not each chunk.
+                # Chunks run in parallel, so summing their runtimes overcounts;
+                # measure from the earliest chunk start of this (job, task) group
+                # and, when over the cap, cancel the whole group (running + queued
```
Yoshi325: This comment fails to mention the inclusion of Not Started and Importing in what it cancels. Probably should just drop the individual identifiers.

hashview/api/routes.py
```diff
@@ -201,14 +205,67 @@ def v1_api_get_admin_settings():
...
+    from the earliest chunk start). 0/None disables the cap."""
+    if not max_hours or max_hours <= 0:
+        return False
+    started = _parent_task_started_at(job_id, task_id)
+    return started is not None and started + timedelta(hours=max_hours) < datetime.now()
+
+
+def _cancel_task_group(job_id, task_id):
+    """Cancel every still-active chunk of a (job, task) group (the running chunk
+    plus any queued/importing siblings), so an over-limit task stops entirely and
```
Yoshi325: docstring fails to mention Not Started as part of what it cancels, probably should drop the individual callouts

migrations/versions/32eb61afc767_add_chunking_and_agent_benchmarks.py
```diff
@@ -0,0 +1,66 @@
+"""Add task chunking + per-agent per-hashtype benchmarks
+
+Adds:
+  - agent_benchmarks table (per-agent, per-hash_type hashcat speed)
+  - job_tasks.chunk_no / job_tasks.chunk_total (a chunk is a JobTasks row)
+  - settings.enabled_chunking / settings.chunk_target_duration
+
```
Yoshi325, line 8: The list above is not exhaustive. There are other chunk_* columns that are added.

hashview/api/routes.py
```diff
@@ -301,11 +373,30 @@ def v1_api_set_agent_heartbeat():
...
+                        if json_response.get('GPU_Count') is not None:
+                            agent.gpu_count = json_response['GPU_Count']
+                        if json_response.get('GPU_Model'):
+                            agent.gpu_model = json_response['GPU_Model']
+                        if json_response.get('Temps'):
+                            agent.gpu_temps = json_response['Temps']
+                    except (ValueError, KeyError, TypeError):
+                        current_app.logger.warning(
+                            'Heartbeat from agent %s had unparseable hc_status; '
+                            'skipping telemetry update.', uuid)
```
Yoshi325: This block is a little concerning. First off, I don't think the comment quite lines up with what is happening here. Ideally the try/except would be wrapped tighter (around only the json.loads) and "Speed #" would be another call to get.

If the goal really is valid parsing of an hc_status dict, we should look into a more robust serialization/deserialization mechanism such as marshmallow.

hashview/agents/routes.py
```diff
@@ -78,10 +130,36 @@ def agents_list():
...
+    in-use hash types on its next idle check-in, refreshing the data the chunk
+    planner sizes from. Admin-only + CSRF-protected.
+    """
+    if not current_user.admin:
+        abort(403)
+    if not FlaskForm().validate_on_submit():
+        flash('Security check failed (invalid or missing CSRF token).', 'danger')
+        return redirect(url_for('agents.agents_list'))
+
+    AgentBenchmarks.query.delete()
```
Yoshi325: **Audit-log this reset.** `agents_benchmark_all` deletes every stored `AgentBenchmarks` row — a destructive admin action directly analogous to `settings/routes.py`'s `log_event('logs.clear', …)`, which *is* audited. This new route records no audit entry (and `agents/routes.py` doesn't import `log_event` yet). `query.delete()` returns the row count, so capture it and log after the successful commit:

```python
n = AgentBenchmarks.query.delete()
if not try_commit('flush agent benchmarks'):
    flash('Could not reset benchmarks; please try again.', 'danger')
    return redirect(url_for('agents.agents_list'))
log_event('agent.benchmarks.reset', detail=f'rows={n}')
```

hashview/api/routes.py
```diff
@@ -201,14 +205,67 @@ def v1_api_get_admin_settings():
...
+def _task_runtime_exceeded(job_id, task_id, max_hours):
+    """True if the parent task has been running longer than max_hours (wall-clock
+    from the earliest chunk start). 0/None disables the cap."""
+    if not max_hours or max_hours <= 0:
+        return False
+    started = _parent_task_started_at(job_id, task_id)
+    return started is not None and started + timedelta(hours=max_hours) < datetime.now()
+
+
+def _cancel_task_group(job_id, task_id):
```
Yoshi325, line 238: **Log *why* work was cancelled.** The new helper in this PR (and `_cancel_job_active_tasks` at line 255) flips chunks to `Canceled`. The agent logs `"Server canceled this task"`, but the server logs nothing about the cause — so an operator can't tell a runtime-cap cancel from an all-recovered cancel from a manual stop. An `info` line in each makes the reason traceable:

```python
# _cancel_task_group (runtime cap)
current_app.logger.info(
    'Task group (job %s, task %s) exceeded max_runtime_tasks; cancelling active chunks.',
    job_id, task_id)

# _cancel_job_active_tasks (hashfile fully recovered)
current_app.logger.info(
    'Job %s hashfile fully recovered; cancelling remaining tasks.', job_id)
```

hashview/api/routes.py
```diff
@@ -354,6 +500,60 @@ def v1_api_set_agent_heartbeat():
...
+        except (TypeError, ValueError):
+            continue
+        if not slowest_benchmark(m):
+            pending.add(m)
+
+    for mode, speed in results.items():
+        try:
+            mode_i = int(mode)
+            speed_i = int(float(speed))
+        except (TypeError, ValueError):
```
Yoshi325, line 544: **Don't drop malformed agent data silently.** Skipping an unparseable entry rather than failing the batch is right, but with no trace a misbehaving agent is undiagnosable. A `warning` naming the agent and the offending value makes it visible:

```python
except (TypeError, ValueError):
    current_app.logger.warning(
        'Agent %s sent an unparseable benchmark entry %r=%r; skipping.',
        agent.id, mode, speed)
    continue
```

(The `pending` loop just above does the same silent skip — `debug` is fine there.)

hashview/models.py
i128: fun fact, neither. for bench mark its just raw output.

hashview/agents/routes.py
Yoshi325: Change lgtm.

You might consider renaming `test_fmt_speed_units` since that doesn't match the current function, and expanding test coverage for edge cases that aren't currently tested.

```python
def test_fmt_hps_default_places_and_all_units():
    assert fmt_hps(0) == "0 H/s"
    assert fmt_hps(999) == "999 H/s"
    assert fmt_hps(2500) == "2.5 kH/s"          # default places=1
    assert fmt_hps(2_500_000) == "2.5 MH/s"     # MH/s branch
    assert fmt_hps(2.8e12) == "2.8 TH/s"        # TH/s branch
    assert fmt_hps(2.8e15) == "2.8 PH/s"        # PH/s branch
    ```

hashview/agents/routes.py
Yoshi325: the `if not by_agent: return ... # skip the empty-IN() query guard` doesn't currently have test coverage, probably could be deferred to another PR though

hashview/api/routes.py
Yoshi325: the update looks good, but it would be idea to have an additional test that tests the path taken without the "Speed #" key

Yoshi325:
Your changes have resolved most of the threads. But we have a few outstanding because the test coverage could be improved. That said, I think the improved test coverage could be deferred to another Issue+PR, so feel free to resolve those threads if you agree.

PR #292, opened 2026-06-29
fix(analytics): decode $HEX[...] before length/strength analysis (#291)

### LINKED ISSUE #291: Analytics mis-measures $HEX[...] cracked passwords (counts the wrapper, not the decoded length)
### Summary

Cracked plaintexts that aren't valid UTF-8 are stored as hashcat's lossless `$HEX[<hex>]` marker (`bytes_to_text`, `hashview/utils/utils.py:430`, via `hexplain_to_text` at ingest, `hashview/api/routes.py:1519`). The **Analytics** dashboard computes password length and every length/character-class-derived statistic directly on that stored string **without decoding it**, so the `$HEX[...]` wrapper inflates the numbers.

`hashview/analytics/routes.py`:
- `:312` `length_counter[len(pword)]` — length-distribution chart
- `:222` `length = len(plaintext)` — strength score (`>=8/12/16`)
- `:251` `_len_col(len(pword))` — length×class heatmap
- `:326` `min(len(pword), 16)` — complexity histogram
- `_char_classes`/`_mask` also run on the literal `$HEX[...]` text

### Impact

| Password | Stored | Real length | Analytics length |
|---|---|---|---|
| `pässwörd` | `$HEX[70c3a4737377c3b67264]` | 8 | **26** |
| bytes `\xff\xfe` | `$HEX[fffe]` | 2 | **9** |

Non-UTF-8 cracks are pushed into the "16+" length bucket and over-scored as "strong"; top-passwords/masks show the wrapper text.

### Note: Wrapped already does this correctly

`hashview/wrapped/routes.py:32` `_decode_plain()` decodes `$HEX[...]` (UTF-8, latin-1 fallback) before measuring length, with a regression test (`tests/unit/test_wrapped_view.py:107`). Analytics should reuse the same logic.

### Proposed fix

Promote `_decode_plain` to a shared `decode_hex_plain()` in `hashview/utils/utils.py`; use it in both Wrapped and Analytics. In Analytics, decode the recovered corpus once at the source so every chart (length, strength, classes, masks, top passwords) measures the real password.

### Tests

Add `$HEX[...]` cases to `tests/unit/test_analytics_view.py` (committed as `xfail` referencing this issue until the fix lands) plus a unit test for the new shared helper.

bandrel:
Closes #291

## Problem

Non-UTF-8 cracked passwords are stored as hashcat's lossless `$HEX[<hex>]` marker (`bytes_to_text`). The **Analytics** dashboard computed password length and every length/character-class-derived statistic directly on that raw string, so the wrapper inflated the numbers:

| Password | Stored | Real length | Analytics length |
|---|---|---|---|
| `pässwörd` | `$HEX[70c3a4737377c3b67264]` | 8 | **26** |
| bytes `\xff\xfe` | `$HEX[fffe]` | 2 | **9** |

Non-UTF-8 cracks were pushed into the "16+" length bucket, over-scored as "strong", and shown as the raw `$HEX[...]` text in top-passwords/masks. **Wrapped** already decoded correctly (`_decode_plain`), so the two pages disagreed.

## Fix

- Promote Wrapped's `_decode_plain` to a shared `decode_hex_plain()` in `hashview/utils/utils.py` — the inverse of `bytes_to_text` (UTF-8, then latin-1 fallback; plain text passes through; `''` for `None`).
- **Analytics** decodes the recovered corpus once at the source, so every chart (length, strength, character classes, masks, top passwords) measures the real password.
- **Wrapped** delegates to the shared helper (behavior unchanged).
- Exports still serve the verbatim `$HEX[...]` (correct for hashcat re-import).

## Tests

- `tests/unit/test_analytics_view.py`: added `$HEX[...]` cases — committed first as strict `xfail` (red, demonstrating the bug), then un-xfailed by the fix.
- `tests/unit/test_unicode_storage.py`: unit test for `decode_hex_plain` (incl. the `pässwörd` 8-vs-26 case), next to the existing `bytes_to_text` tests.
- Full unit suite: **1181 passed**, 0 failures. ruff-clean on all production files.

### DIFF
--- a/hashview/analytics/routes.py
+++ b/hashview/analytics/routes.py
@@ -20,6 +20,7 @@ from sqlalchemy import func, select
 from werkzeug.utils import secure_filename
 
 from hashview.models import Customers, Hashes, HashfileHashes, Hashfiles, Jobs, Tasks, db
+from hashview.utils.utils import decode_hex_plain
 
 analytics = Blueprint('analytics', __name__)
 
@@ -293,8 +294,13 @@ def get_analytics():
 
     # ---- one recovered (plaintext, username) corpus query feeds every
     #      plaintext-derived chart below ----
-    corpus = (_scoped_hash_query(customer_id, hashfile_id, cracked=True)
-              .with_entities(Hashes.plaintext, HashfileHashes.username).all())
+    # Decode hashcat's `$HEX[...]` marker once here so every plaintext-derived
+    # chart below (length, strength, char-classes, masks, top passwords) measures
+    # the real password rather than the wrapper. Mirrors Wrapped's _decode_plain.
+    corpus = [(decode_hex_plain(plaintext), username)
+              for plaintext, username in
+              _scoped_hash_query(customer_id, hashfile_id, cracked=True)
+              .with_entities(Hashes.plaintext, HashfileHashes.username).all()]
     total_cracked = len(corpus)
 
     freq = Counter()
--- a/hashview/utils/utils.py
+++ b/hashview/utils/utils.py
@@ -440,6 +440,24 @@ def bytes_to_text(raw):
     except UnicodeDecodeError:
         return '$HEX[' + raw.hex() + ']'
 
+def decode_hex_plain(plaintext):
+    """Inverse of the ``$HEX[<hex>]`` marker produced by bytes_to_text: return the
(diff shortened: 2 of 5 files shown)


i128:
looks good. good find and fix

PR #293, opened 2026-06-29
fix(auth): redirect authenticated /login + harden e2e; characterize wordlist file encodings

bandrel:
Work that came out of end-to-end testing the wordlist upload + gzip feature.

## 1. `GET /login` redirects authenticated users (`hashview/users/routes.py`)

An already-authenticated user hitting `/login` got the authenticated app shell with an **empty** content area, because the login form lives in the layout's `content_anon` block (suppressed when logged in). `login_get` now redirects them to `main.home`. Unit test added.

This also fixed a real e2e flake: the first-run setup flow auto-logs-in the admin (`setup/routes.py`), so the `login` fixture then landed on an authenticated `/login` where `get_by_label("Email")` resolved to the hidden "Email address" field in the account-settings panel and timed out.

## 2. e2e harness hardening (`tests/conftest.py`, `tests/e2e/test_wordlist_lifecycle.py`)

- `login` fixture short-circuits when already authenticated and targets the login form by `#email`/`#password` (not the ambiguous label).
- The wordlist lifecycle test takes an arbitrary upload via `HASHVIEW_E2E_WORDLIST` (plain text or gzip), defaulting to the bundled example; skips cleanly if the configured path is missing.

## 3. File-encoding characterization tests (`tests/unit/test_wordlist_encoding.py`)

Pin the byte-oriented ingest contract (`ingest_static_wordlist_file`): UTF-8/multibyte, UTF-8 BOM (not stripped), Latin-1, CRLF, UTF-16, and hashcat's `$HEX[...]` candidate format all round-trip verbatim with `\n`-based line counts. Documents the intentional asymmetry where the drop-folder gate (`looks_like_text_or_gz`) rejects UTF-16 (NUL bytes) while UI/API upload accepts it.

## Verification

- Full unit suite: 1179 passed (pre-#291). New module: 17 passing.
- Full e2e suite against a real docker-compose stack (app + MySQL): 28 passed, 5 skipped (missing seed data); the previously-failing first-run path now passes.
- ruff-clean on the production file (`hashview/users/routes.py`).

### DIFF
--- a/hashview/users/routes.py
+++ b/hashview/users/routes.py
@@ -85,6 +85,12 @@ def _azure_enabled():
 def login_get():
     """Function to present login page"""
 
+    # An authenticated user has no login form to fill: the layout only emits the
+    # `content_anon` block (which holds the form) when logged out, so rendering
+    # this page while authenticated yields a blank app shell. Send them home.
+    if current_user.is_authenticated:
+        return redirect(url_for('main.home'))
+
     form = LoginForm()
     return render_template('login.html.j2', title='Login', form=form,
                            azure_enabled=_azure_enabled())
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -137,8 +137,16 @@ def test_user_credentials():
 def login(page, live_server, test_user_credentials):
     def _login():
         page.goto(f"{live_server}/login", wait_until="domcontentloaded")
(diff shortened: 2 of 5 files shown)


i128:
looks great, good find and fix

PR #309, opened 2026-07-02
fix(api): guard None hash in post-import notification sweep (500 on cracked-hash import)

bandrel:
## Problem

`POST /v1/hashes/import/<hash_type>` returns **HTTP 500** even for valid uploads. Observed live:

```
File "hashview/api/routes.py", line 1268, in v1_api_hashes_import
    if hash.cracked:
AttributeError: 'NoneType' object has no attribute 'cracked'
```

The crash is **not** in the import/matching logic (that's guarded with `if record:`). It's in the **"Send Hash Completion Notifications"** sweep that runs *after* a successful import:

```python
hash = Hashes.query.get(hash_notification.hash_id)
if hash.cracked:      # <-- None.cracked when the notification is orphaned
```

If any `HashNotifications` row references a `hash_id` that no longer resolves (deleted/stale hash), `Hashes.query.get()` returns `None` and the whole request 500s — **regardless of what the client submitted**. This is why the recently-merged NTLM import fixes (#305, #307) didn't resolve the 500: they fixed matching, not this sweep.

## Fix

Guard both unguarded sites with `if hash and hash.cracked:` — mirroring the already-guarded third site (~`routes.py:1016`):

- `v1_api_hashes_import` (the import endpoint)
- `v1_api_put_jobtask_crackfile_upload` (the agent crackfile-upload path — same latent bug)

## Test

Adds `test_import_survives_orphaned_hash_notification`: seeds an orphaned `HashNotifications` row, then asserts import returns `200` (and still marks the legitimate hash cracked). Verified it reproduces the exact `AttributeError` at routes.py:1268 without the guard, and passes with it. Full unit suite green.

### DIFF
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -936,19 +936,20 @@ def v1_api_post_jobtask_crackfile_upload(job_task_id):
 
         # Check if hash is cracked
         hash = Hashes.query.get(hash_notification.hash_id)
-        if hash.cracked:
+        if hash:
+            if hash.cracked:
 
-            message += 'You can check the results using the following link: ' + "\n"
-            message += url_for('searches.searches_list', hash_id=hash.id, _external=True)
-            if hash_notification.method == 'email':
-                send_email(user, 'Hashview User Hash Recovered!', message)
-            elif hash_notification.method == 'push':
-                if user.pushover_user_key and user.pushover_app_id:
-                    send_pushover(user, 'Message from Hashview', message)
-            else:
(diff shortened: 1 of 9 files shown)


bandrel:
Superseded — the None-hash guard is already in v0.8.3-dev at hashview/utils/utils.py:374 (`if not hash or not hash.cracked:`). The three inline notification sweeps were refactored into `process_recovered_hash_notifications()` with the guard applied.

PR #312, opened 2026-07-06
test: end-to-end main->dev database migration test

bandrel:
## What

Adds an end-to-end migration test that proves a MySQL database created and populated by the **current `main`** app image upgrades correctly — schema **and** data — when the **dev** app image boots against the same database.

Implements `docs/superpowers/specs/2026-07-02-migration-e2e-test-design.md` (design) and `docs/superpowers/plans/2026-07-06-migration-e2e-test.md` (plan).

## Why this needs real app images (not just Alembic)

`main` has 43 migrations; dev has 57 (14 new). One of the changes is a **data transform that is NOT in the Alembic chain**: migration `a1f7c4e9d2b3` only adds `settings.passwords_decoded=0`; the actual legacy **hex → text/`$HEX[...]`** rewrite of `hashfile_hashes.username` and `hashes.plaintext` runs at **dev app startup** in `decode_legacy_hex_if_needed`. An Alembic-only test would miss it entirely.

## How it works

`tests/run_migration_e2e.sh` orchestrates a compose stack (`docker-compose.migration.yml`):

1. Build `hashview:main` (from `origin/main`) and `hashview:dev` (from this branch).
2. `app-main` boots against a fresh MySQL → creates the main-era schema.
3. Load `tests/migration/seed_main.sql` — legacy-format rows (incl. a **settings row**, without which the backfill never fires) using only columns that exist on main.
4. `app-dev` boots against the **same** DB → applies the 14 migrations + runs the hex backfill. Gated on both alembic head and `passwords_decoded=1`.
5. A second MySQL (`db-fresh`) gets the dev schema built from empty as a parity oracle.
6. `tests/integration/test_migration_e2e.py` (`@pytest.mark.mysql`/`migration`) asserts: alembic at dev head, `passwords_decoded` flip, every hex→text case, DDL-column server-defaults on pre-existing rows, explicit type/charset on the widened+utf8mb4 columns, and **full `information_schema` parity** (columns/charset/collation/indexes/FKs) vs the fresh build.

Runs on push/PR via `.github/workflows/migration-e2e.yml`.

## Verification

Ran the full stack locally: **12/12 verifier tests pass** end to end.

## Notes

- Test-only change; no application behavior modified, so no version bump (that belongs to the v0.8.3 release when dev merges to main).
- The `_decode_hex_column` overflow guard is intentionally out of scope — both columns are `VARCHAR(256)`, so a decoded value can never exceed the width (unreachable end-to-end).

### DIFF
--- a/.github/workflows/migration-e2e.yml
+++ b/.github/workflows/migration-e2e.yml
@@ -11,8 +11,9 @@ jobs:
   migration-e2e:
     runs-on: ubuntu-latest
     # Builds two app images (hashview:main from origin/main + hashview:dev from
-    # the checkout) and cycles a MySQL compose stack, so it needs headroom.
-    timeout-minutes: 25
+    # the checkout) and cycles a MySQL compose stack. Two cold docker builds +
+    # two MySQL boots + a venv install need headroom on a fresh runner.
+    timeout-minutes: 35
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/migration-e2e.yml
+++ b/.github/workflows/migration-e2e.yml
@@ -1,0 +1,34 @@
+name: Migration E2E
+
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  migration-e2e:
+    runs-on: ubuntu-latest
+    # Builds two app images (hashview:main from origin/main + hashview:dev from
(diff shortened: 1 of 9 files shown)


bandrel:
Superseded by #313 — this PR auto-closed when the branch `work/off-dev` was renamed to `tests/db-migration-e2e`. Same commits, continued in #313.

PR #339, opened 2026-07-17
merge 17 commits from main into v0.8.3-dev 

bandrel:


hashview/utils/utils.py
i128, line 1: This change re-introduces the manual rc4/ntlm generation. which we dont want.

bandrel:
Following up on the `utils.py` comment.

**What the code is.** The `main` side adds `_md4_pure` (a pure-Python RFC 1320 MD4) and `ntlm_hash_hex` (MD4 over UTF-16LE). There is no hand-rolled RC4 anywhere — the only RC4 strings in the repo are hashcat mode labels in `hashcat_modes.py`.

**Why it is there.** `hashlib.new("md4")` raises `ValueError: unsupported hash type md4` on OpenSSL 3 builds without the legacy provider (verified locally), so on those hosts the verifier would fail without a fallback.

**Where it runs.** Only from the cracked-hash import verifiers (`CRACKED_HASH_VERIFIERS`, used by `/v1/hashes/import/<hash_type>`, `api/routes.py:1893`): the server recomputes a hash from a submitted plaintext and compares it to the ciphertext already in the DB before marking anything cracked. That is verify-on-import, which you confirmed on #356 you want to keep — not server-side hash generation, and no keys or secrets are involved.

**Status.** This same code already landed on `v0.8.3-dev` independently in `8f12c6f` ("feat(api): verify-only cracked-hash import for all re-hashable types"), so this merge is not what introduces it. Removing it would be a change to `v0.8.3-dev` itself, not a fix to this PR.

**On replacing it with a library.** I looked at swapping `_md4_pure` for `pycryptodome`'s `Crypto.Hash.MD4` and don't think it's worth it:

- `Crypto.Hash.MD4`'s own docs say "This algorithm is insecure. Do not use it for new designs," and pycryptodome documents these primitives as existing only for backward compatibility. Bandit and pip-audit flag MD4 usage either way, so the dependency removes no finding.
- It trades a 106-line pure-Python function for a compiled C dependency: per-platform wheels, added supply-chain surface, and extra CI/Docker weight.
- Correctness is already pinned by RFC 1320 test vectors (`tests/unit/test_api_endpoints.py:319`).
- Performance is not a factor: ~65k hashes/sec in pure Python here, against one hash per submitted plaintext.

MD4 is a frozen 1990 spec, so there is no ongoing maintenance risk. My preference is to keep `_md4_pure` as the OpenSSL-3 fallback. If you still want the hand-rolled implementation out of the tree, tell me and I'll do it on its own branch — but I'd rather not add a dependency for it.

Separately, #339 has real conflicts in `utils.py`, `api/routes.py`, `searches/routes.py`, `CHANGELOG.md` and several tests. I'll resolve those once you've weighed in, so I'm not resolving `utils.py` twice.

bandrel:
Investigated the conflicts: all 8 commits unique to `main` (excluding merge commits) are already superseded by equivalent or improved fixes on `v0.8.3-dev`:

- `b541bc7` CWE-78 command-injection fix in wordlist/rules download → already fixed in-process (gzip, no `os.system`), plus `save_file` random-hex naming
- `9f04bbb`/`6207e4b` NTLM cracked-hash import fixes → superseded by the generic `get_cracked_hash_verifier` rewrite (case-insensitive, correct type handling)
- `11e556f` `/v1/not_authorized` redirects → present
- `1bb6878` `GET /v1/hashfiles/hash_type/<hash_type>` → present
- `b020fb1` `JobNotifications` real columns on `/v1/jobs/add` → present, improved
- `0507a88` search export `str_io`/`strIO` fix → present
- `698449e` hash-import notification null-guard → present via `process_recovered_hash_notifications`
- `5139222` pytest version bump → already at the same version

Verified by test-merging `v0.8.3-dev` into `main` locally and running the full unit/security/agent suite (1492 passed) after resolving conflicts in favor of `v0.8.3-dev`. Nothing from `main` needs to land on `v0.8.3-dev`, so closing rather than force-merging through main's protected branch and an unrelated pre-push hook false-positive issue.

PR #340, opened 2026-07-18
fix(api): return 404 instead of 500 on missing records in stream/job endpoints (#218)

### LINKED ISSUE #218: API: null-deref 500s from missing existence checks (getHashType, jobTasks, uploadCrackFile, search)
Several handlers dereference query results without a `None` check, returning an opaque HTTP 200 wrapping a 500 instead of a clean 404/400.

- `/v1/getHashType/<id>` (`hashview/api/routes.py:1345`) — `HashfileHashes...first()` → `.hash_id` crashes for an unknown/empty hashfile.
- `/v1/jobTasks/<id>` (`routes.py:621`) — authorized for both user+agent, but `agent = Agents...first()` is `None` for a *user* caller, then `agent.id` crashes.
- `/v1/uploadCrackFile/<job_task_id>` (`routes.py:1244`) — `JobTasks.query.get()` → `.job_id` chain crashes on a bad id.
- `/v1/search` (`routes.py:1395`) — `search_json['hash']` is a raw key access; a body without `hash` raises `KeyError` → 500. Should use `.get()`.

## Suggested fix
Add `None` checks after each lookup and return proper 404/400 JSON responses.

MohammadYusif:
## Summary

Several `/v1` API handlers dereferenced query results without a `None` check, so an unknown/missing id produced an `AttributeError` (or a `KeyError` for a body missing the expected key). Flask turned those unhandled exceptions into an opaque HTTP 500 wrapped in a 200, instead of a clean 4xx. Fixes #218.

## Changes

- `hashview/api/routes.py`
  - `GET /v1/getHashType/<hashfile_id>` — guard the `HashfileHashes(...).first()` and `Hashes.query.get(...)` lookups; return `404` when the hashfile/hash is unknown before touching `.hash_id` / `.hash_type`.
  - `GET /v1/jobTasks/<job_task_id>` — the route authorizes user callers too, but a user has no `Agents` row, so `agent.id` crashed on `None`; return `404` when there is no matching agent.
  - `POST /v1/uploadCrackFile/<job_task_id>` — guard each step of the `JobTasks → Jobs → Hashfiles → HashfileHashes → Hashes` lookup chain and return `404` on the first missing record.
  - `POST /v1/search` — use `search_json.get('hash')` instead of `search_json['hash']` so a body without a `hash` key falls through to the existing "Invalid Search" response instead of raising `KeyError`.
- `tests/unit/test_api_issue_218.py` — regression tests asserting each endpoint returns a clean 4xx / handled response for the missing-record cases, plus a happy-path `getHashType` test proving the success response is unchanged.

All new tests fail on `main` (500s) and pass with this change. The error-response shape follows the existing `jsonify({'status': ..., 'type': 'Error', 'msg': ...})` convention already used elsewhere in the blueprint, now returned with a real HTTP 404 status.

Fixes #218

### DIFF
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -933,6 +933,12 @@ def v1_api_get_queue_assignment(job_task_id):
 
     # Get agent id from UUID
     agent = Agents.query.filter_by(uuid=request.cookies.get('uuid')).first()
+    if not agent:
+        return jsonify({
+            'status': 404,
+            'type': 'Error',
+            'msg': 'Agent not found'
+        }), 404
     job_task = JobTasks.query.filter_by(agent_id=agent.id).first()
 
     message = {
@@ -1659,18 +1665,48 @@ def v1_api_post_jobtask_crackfile_upload(job_task_id):
 
     # Get Hashtype from job_task_id
     job_task = JobTasks.query.get(job_task_id)
+    if not job_task:
+        return jsonify({
+            'status': 404,
+            'type': 'Error',
+            'msg': 'Job task not found'
+        }), 404
 
     # Get Job from job_task
     job = Jobs.query.get(job_task.job_id)
+    if not job:
+        return jsonify({
+            'status': 404,
+            'type': 'Error',
+            'msg': 'Job not found'
+        }), 404
 
     # Get hashfile from job
     hashfile = Hashfiles.query.get(job.hashfile_id)
+    if not hashfile:
+        return jsonify({
+            'status': 404,
+            'type': 'Error',
+            'msg': 'Hashfile not found'
+        }), 404
 
     # Get hashfilehashes from hashfile
     hashfilehashes = HashfileHashes.query.filter_by(hashfile_id=hashfile.id).first()
+    if not hashfilehashes:
+        return jsonify({
+            'status': 404,
+            'type': 'Error',
+            'msg': 'Hashfile hashes not found'
+        }), 404
 
     # Get single hash
     single_hash = Hashes.query.get(hashfilehashes.hash_id)
+    if not single_hash:
+        return jsonify({
+            'status': 404,
+            'type': 'Error',
+            'msg': 'Hash not found'
+        }), 404
 
     hash_type = single_hash.hash_type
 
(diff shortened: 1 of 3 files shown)


i128:
Please re-submit as a pr to v0.8.3-dev branch instead.

bandrel:
Thanks @MohammadYusif — the guards are right and this is getting merged. Two things I did directly on the branch, both explained below so nothing here is a surprise.

### 1. Rebased onto `v0.8.3-dev` and retargeted the PR

This was opened against `main`, but `main` is our release branch — active work lands on `v0.8.3-dev` first. That mattered more than usual here: **`main`'s CI runs only E2E and pylint, no unit-test job**, so the green check on this PR never actually executed your `tests/unit/test_api_issue_218.py`. Retargeting is what puts these tests under a gate.

The branch is now `v0.8.3-dev` + your fix + tests (force-pushed). Your original commits are still recoverable if you need them: `407bb49` (the fix) and `dfe9b51` (tests) — the PR timeline retains both.

Two things came out of the rebase worth knowing:

- **The `/v1/search` hunk was dropped**, resolved in favour of dev. That handler was rewritten on `v0.8.3-dev` (it now supports `hash`/`plaintext`/`username` and uses `request.get_json(silent=True)`), and it already does `search_json.get('hash')` — your change had landed independently. Every other guard applied cleanly, and all eight are on the branch.
- **Test cookies needed `domain="localhost.test"`.** Dev's test app sets `SERVER_NAME`, so Werkzeug 3.x won't send a cookie scoped to the default domain and every request 302'd to `/v1/not_authorized`. Matches the pattern in `tests/unit/test_api_endpoints.py`.

### 2. Added tests, and tightened one of yours

Five of your eight guards were unexercised, so I pushed coverage for them: the `uploadCrackFile` chain (`job` → `hashfile` → `hashfilehashes` → `hash`, each reached with an orphaned row — those are plain integer columns, not enforced FKs, and deletes have historically left orphans behind), the second hop in `getHashType`, and a `'hash'` key that's present but empty.

I also changed one of your assertions, which is the part I'd want a second opinion on if you disagree:

```python
# before
assert resp.status_code in (200, 404)
# after
assert resp.status_code == 404
```

`in (200, 404)` passes whether or not the guard fires, so it documents the intent but doesn't pin the fix — the pre-fix code raised `AttributeError` → 500, and a future regression that returned 200 with a null body would also slip through. It's now an exact 404, plus a new test asserting the *agent* caller still gets a 200, so the guard can't be "fixed" by breaking the path it exists for.

I verified each test fails without the fix: stripping all eight guards and reverting `.get('hash')` makes all nine 404 tests fail with the original `AttributeError`/`KeyError` 500s.

### 3. Retired three strict xfails

`v0.8.3-dev` had `#218` documented in `tests/unit/test_api_issues_xfail.py` as `xfail(strict=True)` — so your fix turned them XPASS, which is a hard failure by design. That file's convention is to drop the marker and fold the test into the normal suite on XPASS; since your new file covers the same endpoints and more, I deleted the three and left a pointer.

Full unit suite on the rebased branch: **1527 passed, 2 skipped, 36 xfailed**. Nice catch on this one — the orphaned-row cases in particular are the kind of thing that only shows up in a long-lived install.

i128:
Awesome, thanks for your help on this!

Issue #297, opened 2026-06-30
Security (HIGH): command injection via task mask/rule fields into agent shell


bandrel:
## Summary

Free-form task fields flow unquoted/unescaped into a hashcat command string that the agent executes via `subprocess.Popen(command, shell=True)`. Shell metacharacters in those fields therefore execute arbitrary commands **on the agent host**.

## Details

- `Tasks.hc_mask`, `Tasks.j_rule`, `Tasks.k_rule` are free-form `StringField`s with **no validators** (`hashview/tasks/forms.py`).
- `build_hashcat_command` (`hashview/utils/utils.py`, ~line 785+) concatenates them into the command string:
  - **unquoted** for the mask (attack modes 3/6/7), and
  - **single-quoted but unescaped** for the j/k rules (mode 1) — so a `'` breaks out of the quoting.
- The resulting string is stored on `JobTasks.command` and run by the agent via `subprocess.Popen(command, shell=True)` (`install/hashview-agent/hashview-agent.py`).

Net effect: payloads such as `; whoami #`, `$(...)`, backticks, `&&`, `|`, newlines (mask) and single-quote break-out (j/k rules) execute on the agent host.

## Severity

HIGH — remote command execution on agent hosts, reachable by any user who can create/edit a task.

## Reproduction / proof

Demonstrated by strict-xfail regression tests in `tests/security/test_security_hardening.py`:
- `test_mask_field_shell_injection`
- `test_combinator_rule_shell_injection`

These feed shell metacharacters through the task fields into `build_hashcat_command` and assert the payload does **not** reach the command raw. They are currently `xfail(strict=True)` — when this is fixed, they flip to visible passes and the `xfail` markers should be removed.

## Suggested remediation

- Build the hashcat invocation as an argv list and run it **without** `shell=True` on the agent, OR
- Strictly validate/escape `hc_mask`, `j_rule`, `k_rule` (e.g. `shlex.quote`, allowlist charset) before they enter the command string.

Found during test-architecture hardening (PR #296).

bandrel:
Regression test added in **#299** (`tests/security/test_command_injection_xfail.py`): 24 strict xfails proving the injection via `hc_mask`/`j_rule`/`k_rule` → `build_hashcat_command` → agent `shell=True`. When this is fixed, the strict xfail flips to a visible pass and the `xfail` markers should be removed.

PR #342, opened 2026-07-21
fix(security): prevent command injection via task mask/rules on agent…

BirkeP:
## Summary
- Fixes CWE-78 command injection where task mask / combinator rules were concatenated into a shell command executed by hashview-agent with `shell=True`.
- Adds form + server-side validation and `shlex.quote()` for user-controlled hashcat fragments.
- Agent runs hashcat with `shell=False` and streams status output in Python instead of `| tee`.

## Impact
Any authenticated user who can create tasks and queue jobs on an agent could execute arbitrary commands on the agent host.

## Related
- Complements PR #337 (download path / os.system hardening on the server).

## Test plan
- [ ] `pytest tests/unit/test_security_agent_task_cmdi.py -v`
- [ ] `pytest tests/unit/test_security_filename_command_injection.py -v`
- [ ] Create a normal mask-mode task (`?l?l?l?d?d`) and confirm job still queues
- [ ] Confirm malicious mask `?a;touch /tmp/x;#` is rejected at task create
- [ ] Deploy patched agent; confirm hashcat job runs and status file updates

### DIFF
--- a/hashview/tasks/forms.py
+++ b/hashview/tasks/forms.py
@@ -3,6 +3,7 @@ from flask_wtf import FlaskForm
 from wtforms import StringField, SubmitField, SelectField
 from wtforms.validators import DataRequired, InputRequired, ValidationError
 from hashview.models import Tasks
+from hashview.utils.utils import validate_hashcat_combinator_rule, validate_hashcat_mask
 
 
 class TasksForm(FlaskForm):
@@ -26,3 +27,21 @@ class TasksForm(FlaskForm):
         task = Tasks.query.filter_by(name = name.data).first()
         if task:
             raise ValidationError('That task name is taken. Please choose a different one.')
+
+    def validate_mask(self, field):
+        try:
+            validate_hashcat_mask(field.data)
+        except ValueError as exc:
+            raise ValidationError(str(exc)) from exc
+
+    def validate_j_rule(self, field):
+        try:
+            validate_hashcat_combinator_rule(field.data)
+        except ValueError as exc:
+            raise ValidationError(str(exc)) from exc
+
+    def validate_k_rule(self, field):
+        try:
+            validate_hashcat_combinator_rule(field.data)
+        except ValueError as exc:
+            raise ValidationError(str(exc)) from exc
--- a/hashview/utils/utils.py
+++ b/hashview/utils/utils.py
@@ -3,6 +3,7 @@ import os
 import secrets
 import hashlib
 import re
+import shlex
 import struct
 import binascii
 import _md5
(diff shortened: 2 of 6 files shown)


i128:
I think this might already be resolved, but please check v0.8.3-dev and pr to that branch instead of main.

bandrel:
Thanks for finding and reporting this, @BirkeP — the vulnerability was real and your report helped drive the fix.

To close the loop: the server-side part of this issue has already landed on `v0.8.3-dev`. `build_hashcat_command` now returns an argv list (not a shell string), so mask and rule values become literal subprocess arguments — shell metacharacters in them cannot execute. The remaining agent-side sink — the `run_command(shell=True)` helper used during rule sync, where the server-supplied filename was interpolated into a shell command line — is addressed by #366 ("fix(agent): stop shelling out during rule sync; bound the server-supplied filename"), which removes `run_command` entirely and replaces the `gunzip`/`mv` shellouts with in-process decompression. #365 ("test(security): make the #297 command-injection tests argv-aware") converts the existing #297 tests into explicit argv-aware regression tests.

Both #365 and #366 are pending review against `v0.8.3-dev`. If there is any part of your diff you feel is not covered by those two PRs, please let us know — we want to make sure nothing is missed.

BirkeP:
Thanks for the detailed follow-up and for the kind words. It sounds like the remaining parts are covered by #365 and #366, so I don’t have anything further to add. I appreciate you taking the report seriously and addressing it so thoroughly.

bandrel:
Thanks for this @BirkeP — the analysis was right, and the fix is sound. I'm closing it as superseded rather than merging, because both halves of it have since landed on `v0.8.3-dev` (our integration branch) by a different route. Details, so it's clear this isn't being dropped on the floor:

**Server side.** `build_hashcat_command` on `v0.8.3-dev` now returns an **argv list** instead of a shell string ([utils.py](https://github.com/hashview/hashview/blob/v0.8.3-dev/hashview/utils/utils.py) — `def build_hashcat_command(job_id, task_id, chunk=None, job_task_id=None)`). `hc_mask`, `j_rule` and `k_rule` are appended as literal list elements, so there is no shell string for a metacharacter to break out of and nothing to quote or screen. That's strictly stronger than validate-and-`shlex.quote` — it removes the sink rather than filtering what reaches it.

**Agent side.** #366 took the agent further than this PR does: it deleted `run_command` outright (rather than making it `shell=False`), decompresses with `gzip` in-process, installs with `os.replace`, and — the part this PR is missing — **sanitizes the server-supplied filename to a plain basename** before it becomes a path. As written here, `filename = entry['path'].split('/')[-1]` lets a stored path of `control/rules/..` resolve to `dest = control/rules/..`, and `shutil.move` then drops the temp file in the parent directory. Not shell injection anymore, but still a path the server controls.

**One thing I'd flag as a genuine gap in your approach**, in case it comes up again: `_HASHCAT_MASK_BAD` rejects `$`, which is a legal literal in a hashcat mask — `Pa$$word?d?d` is a mask a user can reasonably write, and `validate_hashcat_mask` would refuse it. (Your rule regex correctly omits `$`, so `-j $-` / `-k $!` still work — I had that backwards in the first version of this comment.) Denylisting shell syntax at the app layer tends to collide with the target language's own syntax like this, which is part of why we went with argv.

For the record: while reviewing I pushed a test commit to this branch (4cdae95) covering the parts of your fix that were unverified — the `TasksForm` validators, attackmodes 1/6/7, and the agent-side behaviour. Those tests are how I confirmed the fix worked before concluding it was superseded; they close with the PR and no rebase is needed on your end.

Superseded by #366 and the argv change on `v0.8.3-dev`. Genuinely appreciated the report — please do keep filing these.

PR #345, opened 2026-07-21
feat(wordlists): add dynamic wordlist providers

Yoshi325:
Let admins register external wordlist-generation services under Settings -> Wordlist providers (base URL + Bearer/Basic credentials). Each registered provider becomes a (DYNAMIC) wordlist; when a job uses it, the operator supplies an input in the wizard that hashview submits to the provider's API at crack time, materializing the returned list. Agents never contact the provider -- they download the materialized file from hashview like any other dynamic wordlist.

Modeled on the existing (DYNAMIC) Website Keywords crawler:
- new WordlistProviders table; Wordlists.provider_id FK; Jobs.provider_input
- provider HTTP client (bearer/basic auth, /health test, streamed /generate, atomic materialize, never-raises no-op on failure) mirroring utils/crawler.py
- update_dynamic_wordlist() branches on provider_id, leaving the crawler and DB-derived generators unchanged
- Settings CRUD/test UI (write-only secret; provider_secret added to the API serialization denylist so it never reaches agents)
- conditional job-wizard step (jobs_provider_input) chained website -> provider_input -> summary, self-skipping when unused
- guarded, idempotent Alembic migration; docs/wordlist-providers.md contract

Tests:
- unit: provider HTTP client (auth/streaming/no-op-on-failure), the update_dynamic_wordlist provider branch + backward-compat, settings CRUD routes + admin gating, and secret redaction
- rendered-HTML UI (Flask test client): the Settings providers tab + add form, a listed provider that never leaks its secret, the conditional wizard step (and its skip), and the wordlists provider badge
- browser e2e (Playwright, tests/e2e conventions): add a provider through the Settings tab and confirm it lists without exposing the secret then delete it; and confirm a provider is a (DYNAMIC) option in the task wordlist dropdown


i128:
Do we have any example 3rd party services we could 1) test with 2) include in a document for users to understand how to use this feature?

Yoshi325:
> Do we have any example 3rd party services we could 1) test with 2) include in a document for users to understand how to use this feature?

We have https://github.com/hashview/hashview/pull/345/changes/cdc71a07d451ee9e0a4b5b621f78d014a9cb1d97#diff-49b17e4a6c9dbb11d964e0148e249cb80c7b1233419ce6db70796643908e7279R68

But we can improve it if needed.

Yoshi325:
There is a bug in the web scraping dynamic wordlist this was modeled after. We need to decide if that bug should be fixed separately, or in this PR.

bandrel:
Heads up @Yoshi325 — I rebased this branch onto the latest `v0.8.3-dev` to resolve conflicts (your changes were kept as-is). Two mechanical fixes were needed on top:

- Re-parented migration `f1a2b3c4d5e6` onto `d4e7a1c9f2b8` (the theme migration merged since), so Alembic has a single head again.
- Updated the hardcoded `DEV_HEAD` in `tests/run_migration_e2e.sh`, `tests/integration/test_migration_e2e.py`, and `tests/unit/test_migration_drift_idempotency.py` to your new head.

Since this was a force-push, run `git fetch && git reset --hard origin/add-dynamic-wordlist-providers` (or `git pull --rebase`) before making further changes locally. All 1548 non-e2e tests pass on the rebased branch.

PR #348, opened 2026-07-21
ci: bump actions to Node 24 runtime majors

bandrel:
## Summary
Clears the **Node 20 deprecation warnings** on the GitHub runners. GitHub is retiring the Node 20 action runtime (→ Node 24); every `actions/*` step was pinned to a node20 (or node16) major, so each emitted the deprecation notice.

Each action is bumped to its **first Node 24 major** (confirmed by reading each version's `action.yml` `runs.using`):

| Action | Before | After | Note |
|---|---|---|---|
| `actions/checkout` | v4 | **v5** | v4 = node20 |
| `actions/setup-python` | v5 (v3 in `pylint.yml`) | **v6** | v5 = node20, v3 = node16 |
| `actions/upload-artifact` | v4 | **v6** | v5 is still node20 |
| `actions/download-artifact` | v4 | **v7** | v5/v6 are still node20 |

Kept as floating **major tags** (not SHA-pinned), consistent with the existing style.

## Notes
- The artifact actions can't just go `v4→v5`; their first node24 majors are v6 (upload) / v7 (download). Both remain on the v4+ artifact backend, so the upload→download handoff in `unit-tests.yml` still interoperates.
- Second commit adds pre-push scanner allowlist entries for CI-only false positives (test creds, loopback URIs, SecLists checksum/URL, the `${GH_TOKEN}` badge-push URL, "MariaDB"/"regardless" keyword substrings) that surface because editing a workflow makes the scanner read the whole file.

## Verification
- All 25 `actions/*` pins updated across 8 workflow files.
- Every workflow re-parses as valid YAML.
- CI on this PR will itself confirm the warnings are gone.

### DIFF
--- a/.githooks/safe-allowlist.txt
+++ b/.githooks/safe-allowlist.txt
@@ -33,3 +33,23 @@ regardless
 password="hashed-pw"
 api_key="user-api-key-admin"
 src_ip="127\.0\.0\.1"
+
+# --- CI workflow false positives (.github/workflows/*) ---
+# The scanner reads whole changed files, so editing a workflow surfaces its
+# CI-only test values, template variables and comment substrings. None are
+# real secrets.
+
+# "iadb" keyword inside "MariaDB" (job name, image tag, service env, comments).
+mariadb
+# Loopback address in CI service URIs / base URLs.
+127\.0\.0\.1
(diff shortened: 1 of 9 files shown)


i128:
Please resolve conflict before merging

PR #354, opened 2026-07-23
feat(wordlists): length-bucketed dynamic recovered-password wordlists (#353)

### LINKED ISSUE #353: Dynamic wordlists split by password length (dynamic-all, bucketed)
## Summary

Add a family of **dynamic wordlists split by password length**, derived from the same source as the existing `(DYNAMIC) All Recovered Passwords` (aka `dynamic-all`) wordlist, but bucketed so each list contains only recovered plaintexts of a specific length.

## Motivation

`(DYNAMIC) All Recovered Passwords` builds a single flat list of every distinct cracked plaintext (`Hashes.plaintext` where `cracked=True`). When attacking password policies with known minimum lengths, or when running length-constrained mask/hybrid attacks, feeding hashcat the *entire* recovered corpus wastes candidates on passwords that can't possibly match. Length-bucketed dynamic wordlists let an operator target only the relevant candidates, improving crack rate per candidate and cutting runtime.

## Proposed behavior

Use a **fixed set of buckets** (decision made — no on-the-fly generation). Lengths `0–5` are lumped into a single combined bucket; each length from 6 to 8 gets its own bucket, with a `9+` catch-all for anything longer:

- `(DYNAMIC) Recovered Passwords (length 0-5)`
- `(DYNAMIC) Recovered Passwords (length 6)`
- `(DYNAMIC) Recovered Passwords (length 7)`
- `(DYNAMIC) Recovered Passwords (length 8)`
- `(DYNAMIC) Recovered Passwords (length 9+)`

Each is a `type='dynamic'` wordlist that, when regenerated, writes only the distinct recovered plaintexts falling in its length range. The bucket set is a seeding constant (`_PASSWORD_LENGTH_BUCKETS`) and easy to adjust later. Empty buckets are expected and fine — they seed like any other and simply produce an empty list until matching plaintexts exist.

## Implementation notes

Existing machinery this builds on (from code research):

- **Seeding** — `_DYNAMIC_WORDLISTS` tuple in `hashview/setup/__init__.py` (~L265) and mirrored in `hashview.py` (~L135). Add the fixed bucket entries here (a small loop/constant generating the tuple keeps it maintainable).
- **Generation dispatcher** — `update_dynamic_wordlist()` in `hashview/utils/utils.py` (~L737-798) routes by name substring (`'Passwords'`, `'Usernames'`, ...). Extend it to parse the length range from the wordlist name and filter the query:
  ```python
  # MySQL: CHAR_LENGTH counts characters, not bytes
  q = Hashes.query.filter_by(cracked=True)
  # combined low bucket:   func.char_length(...) <= 4
  # exact bucket n:        func.char_length(...) == n
  # catch-all high bucket: func.char_length(...) >= 9
  q.distinct('plaintext').with_entities(Hashes.plaintext)
  ```
  `$HEX[...]` plaintexts are decoded **only to measure their true byte length** for bucketing; the value is **stored in the wordlist in its original `$HEX[...]` form**.
- **Model** — no schema change needed; reuses `Wordlists` (`type='dynamic'`, uncompressed `.txt` on disk) and `Hashes.plaintext` / `Hashes.cracked`.
- **Regeneration** — same on-demand paths as today: UI update button (`hashview/wordlists/routes.py` ~L245) and agent pre-task refresh (`hashview/api/routes.py` ~L803). No new scheduler work required.
- **Tests** — extend `tests/unit/test_update_dynamic_wordlist.py` with the length-filter contract: combined `0-4` bucket, an exact-length bucket, the `9+` catch-all, `$HEX[...]` handling, and empty-bucket behavior.

## Acceptance criteria

- [ ] Fixed set of length buckets is seeded: combined `0-5`, one each for 6/7/8, and a `9+` catch-all.
- [ ] Regenerating a bucket writes only distinct recovered plaintexts in that length range.
- [ ] `$HEX[...]`-encoded plaintexts are bucketed by decoded byte length but stored verbatim in `$HEX[...]` form.
- [ ] Length is computed correctly on the target DB (character count, not bytes).
- [ ] Empty buckets seed and regenerate cleanly (produce an empty list, no error).
- [ ] Unit tests cover the combined, exact, and catch-all dispatcher paths.

## Task references (key requirement)

An operator must be able to create a task that references a specific length bucket directly, e.g. a task using `(DYNAMIC) Recovered Passwords (length 8)`.

This is satisfied for free by the fixed-bucket design: `Tasks.wl_id` (`hashview/models.py:325`) stores an integer FK to `Wordlists.id`, and each seeded bucket has a stable, permanent id. The agent's pre-task refresh (`hashview/api/routes.py` ~L803) already regenerates any dynamic `wl_id` before the task runs, so the bucket is fresh at task time. No `Tasks` model or task-creation changes are required.

Note: this stable-id guarantee is precisely why fixed buckets are required over on-the-fly generation — a saved task's `wl_id` would dangle if the bucket row could be created/destroyed based on current DB contents.

- [ ] A task can be created/saved referencing an individual length bucket (e.g. length 8) by `wl_id`, and it survives even when that bucket is currently empty.

bandrel:
## What

Implements **length-bucketed dynamic wordlists** (closes #353). Alongside `(DYNAMIC) All Recovered Passwords`, seed a fixed set of buckets that split the recovered-password corpus by plaintext length:

- `(DYNAMIC) Recovered Passwords (length 0-5)` — `len <= 5` (combined)
- `(DYNAMIC) Recovered Passwords (length 6)` / `(length 7)` / `(length 8)` — exact length
- `(DYNAMIC) Recovered Passwords (length 9+)` — `len >= 9` (catch-all)

Each is a normal `type='dynamic'` wordlist with a **stable id**, so a task can reference an individual bucket via `Tasks.wl_id` and the agent's pre-task refresh regenerates it before use.

## How

- `update_dynamic_wordlist()` parses the `(length ...)` token from the wordlist name into a predicate (`a-b` → range, `N+` → catch-all, `N` → exact) and filters the recovered plaintexts. **No token → unfiltered**, so the plain "All Recovered Passwords" list is unchanged.
- `$HEX[...]` plaintexts are **decoded to raw bytes first**, then bucketed by decoded length and written as the decoded value. Because decoded bytes may not be valid UTF-8, the passwords branch now writes in **binary mode**.
- Buckets are defined once in `dynamic_password_length_wordlists()` and seeded from both setup paths (`hashview/setup/__init__.py` and `hashview.py`), idempotent per name so existing installs pick them up on next startup.

## Commits

1. `test(wordlists): spec length-bucketed dynamic wordlists as strict xfails` — TDD RED, executable spec
2. `feat(wordlists): length-bucketed dynamic recovered-password wordlists` — GREEN implementation; flips the spec tests to passing

## Tests

```
tests/unit/test_dynamic_wordlist_length_buckets.py   7 passed
tests/unit/test_update_dynamic_wordlist.py           5 passed  (no regression)
tests/unit/test_website_keywords_wordlist.py        12 passed  (seeding path)
```

Coverage: exact bucket, combined `0-5` (boundary 5 in / 6 out), `9+` catch-all, `$HEX` decoded into correct bucket, `$HEX` non-UTF-8 written raw, empty bucket → valid empty file, and a regression guard that the unbucketed list still writes everything.

## No schema change

Reuses `Wordlists` and `Hashes`; no migration required.

Closes #353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ Notable changes will be documented here
 - On-disk JSON audit log of authentication and create/update/delete events, plus captured server errors, viewable and clearable from Settings
 
 **Length-Bucketed Dynamic Wordlists**
-- New `(DYNAMIC) Recovered Passwords (length ...)` wordlists split the recovered-password corpus by plaintext length into a fixed set of buckets — `0-5`, `6`, `7`, `8`, and a `9+` catch-all — so tasks can target only candidates of a relevant length. `$HEX[...]` plaintexts are decoded first, then bucketed by their decoded byte length and written as the decoded value (raw bytes when not valid UTF-8).
+- New `(DYNAMIC) Recovered Passwords (length ...)` wordlists split the recovered-password corpus by plaintext length into a fixed set of buckets — `0-5`, `6`, `7`, `8`, and a `9+` catch-all — so tasks can target only candidates of a relevant length. `$HEX[...]` plaintexts are decoded only to measure their true byte length for bucketing; they are stored in the wordlist in their original `$HEX[...]` form.
 
 **API Documentation**
 - The `/v1` API is now described by an OpenAPI spec with interactive Swagger UI at `/api/docs`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Notable changes will be documented here
 **Audit & Error Logging**
 - On-disk JSON audit log of authentication and create/update/delete events, plus captured server errors, viewable and clearable from Settings
 
+**Length-Bucketed Dynamic Wordlists**
+- New `(DYNAMIC) Recovered Passwords (length ...)` wordlists split the recovered-password corpus by plaintext length into a fixed set of buckets — `0-5`, `6`, `7`, `8`, and a `9+` catch-all — so tasks can target only candidates of a relevant length. `$HEX[...]` plaintexts are decoded first, then bucketed by their decoded byte length and written as the decoded value (raw bytes when not valid UTF-8).
+
 **API Documentation**
 - The `/v1` API is now described by an OpenAPI spec with interactive Swagger UI at `/api/docs`
 - `DELETE /v1/hashfiles/<id>` to remove a hashfile via the API
--- a/hashview.py
+++ b/hashview.py
@@ -140,7 +140,7 @@ def ensure_dynamic_wordlist(db):
     creates: Recovered Passwords, Usernames, Customers, NTLM Hashes.
     """
     from hashview.models import Wordlists
-    from hashview.utils.utils import get_filehash
+    from hashview.utils.utils import dynamic_password_length_wordlists, get_filehash
 
     wanted = [
         ('(DYNAMIC) All Recovered Passwords', 'hashview/control/wordlists/dynamic-all.txt'),
@@ -148,6 +148,8 @@ def ensure_dynamic_wordlist(db):
         ('(DYNAMIC) All Customers',           'hashview/control/wordlists/dynamic-customers.txt'),
         ('(DYNAMIC) All NTLM Hashes',         'hashview/control/wordlists/dynamic-ntlm.txt'),
         ('(DYNAMIC) Website Keywords',        'hashview/control/wordlists/dynamic-website-keywords.txt'),
+        # Recovered passwords split into fixed length buckets (0-5, 6..8, 9+).
+        *dynamic_password_length_wordlists(),
     ]
     added = 0
     for name, path in wanted:
--- a/hashview/setup/__init__.py
+++ b/hashview/setup/__init__.py
@@ -8,6 +8,7 @@ from hashview.models import Hashes, HashfileHashes, Rules, Settings, Tasks, User
 from hashview.utils.utils import (
     bytes_to_text,
     compress_to_gz,
+    dynamic_password_length_wordlists,
     get_filehash,
     get_filesize,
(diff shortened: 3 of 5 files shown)


bandrel:
Follow-up commit `7e714f0` addresses two review-worthy design points:

**1. Single generation function.** All dynamic recovered-password generation now flows through one function — `generate_recovered_password_wordlist(path, min_length=0, max_length=None)` in `hashview/utils/utils.py`. The app passes the length window; `update_dynamic_wordlist()` just parses the `(length ...)` token from the wordlist name into `(min, max)` and delegates. Length-filtering logic lives in exactly one place.

**2. Seeding on install *and* upgrade (even with zero matching passwords).** The buckets are in `_DYNAMIC_WORDLISTS`, so the existing idempotent startup seeder (`default_dynamic_wordlists_need_added` → `add_default_dynamic_wordlists`, run in `create_app`) creates each bucket row + empty file on a fresh install and on upgrade from a v0.8.2 install that predates them — regardless of whether any cracked password currently fits a bucket. This is the same mechanism that rolled the NTLM bucket out to existing installs, so no Alembic data migration is needed (and none is added). New tests lock both the registration and the zero-password upgrade path.

bandrel:
Commit `91d628b`: per review, `$HEX[...]` plaintexts are now **stored verbatim** in the wordlist — the decode step is used *only* to measure true byte length for bucketing/sorting. Since stored plaintext is always valid UTF-8 (real text or the ASCII `$HEX[...]` wrapper), generation returns to UTF-8 text mode (no more binary write). Tests updated accordingly.

PR #360, opened 2026-07-24
feat(api): task-viability read endpoints (GET /v1/tasks, GET /v1/agents/benchmark)

bandrel:
## Feature request

External tooling needs to audit the tasks already on a Hashview server for
viability (is a wordlist×rule run worth the GPU time for a given hash type?)
**over the API, without direct database access.** Two pieces of data required
for that are stored but not currently readable via `/v1`:

1. **The list of tasks.** Only `GET /v1/tasks/{id}` (single) exists today, forcing
   an id-crawl. This adds `GET /v1/tasks` to list all tasks in one call.
2. **Agent benchmark performance.** `POST /v1/agents/benchmark` writes per-(agent,
   hash type) speeds to `agent_benchmarks`, but nothing reads them back over HTTP.
   This adds `GET /v1/agents/benchmark` returning performance aggregated to the
   **slowest agent per hash type** (the same basis the chunk planner uses), with an
   optional `?hash_type=<mode>` filter.

## What changed

- `GET /v1/tasks` — list all tasks (metadata). Mirrors `GET /v1/rules`.
- `GET /v1/agents/benchmark` — read-only; returns `{performance: {mode: H/s}}`, or
  `{hash_type, speed}` for a single mode (`null` when never benchmarked). Reuses the
  existing `slowest_benchmark()` helper. **The agent and its benchmark generation are
  untouched** — this only exposes already-stored data.
- Minor hooks maintenance: allowlist the ordinary word "regardless" (false positive
  for the "gard" keyword) and neutralize a vendor keyword in the allowlist example.

## Compatibility & safety

- Both endpoints are **additive** GETs; no existing route (incl. the benchmark POST
  and single-task GET) is modified.
- Auth matches sibling read routes: `is_authorized(user=True, agent=True)`.
- No new tables/migrations; `Tasks` and `AgentBenchmarks` carry no secrets
  (serialization denylist still strips `password`/`api_key`).

## Tests & docs

- New `tests/unit/test_task_viability_endpoints.py` (auth, listing, slowest-per-hash
  aggregation, single-hash + missing, empty table). Full unit suite green
  (1338 passed).
- `hashview/api_docs/openapi.yaml` updated (required by `test_openapi_spec.py`
  route/spec parity); `CHANGELOG.md` updated.

Consumer/design context: audit tool spec lives in the companion repo
(`hashview-task-creator`), `docs/superpowers/specs/2026-07-23-audit-existing-tasks-design.md`.

### DIFF
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -579,6 +579,7 @@ def v1_api_post_agent_benchmark():
 
     return jsonify({'status': 200, 'type': 'message', 'msg': 'OK'})
 
+
 @api.route('/v1/agents/benchmark', methods=['GET'])
 def v1_api_get_agent_benchmarks():
     # Expose already-stored per-(agent, hash type) benchmarks as rig performance.
@@ -598,7 +599,7 @@ def v1_api_get_agent_benchmarks():
                              db.func.min(AgentBenchmarks.speed))
             .group_by(AgentBenchmarks.hash_type).all())
     return jsonify({'status': 200,
-                    'performance': {ht: spd for ht, spd in rows}})
+                    'performance': {str(ht): spd for ht, spd in rows}})
 
 @api.route('/v1/customers', methods=['GET'])
 def v1_api_get_customers():
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -1124,6 +1124,16 @@ def v1_api_post_start_job(job_id):
             'msg': 'Invalid job ID'
         })
 
+# List all tasks
+@api.route('/v1/tasks', methods=['GET'])
+def v1_api_get_tasks():
+    if not is_authorized(user=True, agent=True, request=request):
+        return redirect("/v1/not_authorized")
+
+    update_heartbeat(request.cookies.get('uuid'))
+    tasks = Tasks.query.all()
+    return jsonify({'status': 200, 'tasks': alchemy_to_native(tasks)})
+
 # Provide task info
 @api.route('/v1/tasks/<int:task_id>', methods=['GET'])
 def v1_api_get_task(task_id):
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -585,19 +585,29 @@ def v1_api_get_agent_benchmarks():
     # Expose already-stored per-(agent, hash type) benchmarks as rig performance.
     # Aggregated per hash type to the SLOWEST agent's speed, matching how chunk
     # sizing works (slowest_benchmark). Read-only; touches no agent code.
-    if not is_authorized(user=True, agent=True, request=request):
+    #
+    # Consumers are humans/dashboards, not agents: an agent has no use for the
+    # fleet-wide view, so agent keys are rejected and no heartbeat is recorded
+    # (a heartbeat would misreport this caller as a live agent).
+    if not is_authorized(user=True, agent=False, request=request):
         return redirect("/v1/not_authorized")
-
-    update_heartbeat(request.cookies.get('uuid'))
 
     hash_type = request.args.get('hash_type', type=int)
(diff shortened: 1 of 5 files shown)


hashview/api/routes.py
```diff
@@ -579,6 +579,28 @@ def v1_api_post_agent_benchmark():
...
+
+@api.route('/v1/agents/benchmark', methods=['GET'])
+def v1_api_get_agent_benchmarks():
+    # Expose already-stored per-(agent, hash type) benchmarks as rig performance.
+    # Aggregated per hash type to the SLOWEST agent's speed, matching how chunk
+    # sizing works (slowest_benchmark). Read-only; touches no agent code.
+    if not is_authorized(user=True, agent=True, request=request):
+        return redirect("/v1/not_authorized")
+
+    update_heartbeat(request.cookies.get('uuid'))
```
i128: you dont need to update the heartbeat if you're not an agent
agents also dont need this information, so agent=False

hashview/api/routes.py
```diff
@@ -579,6 +579,28 @@ def v1_api_post_agent_benchmark():
...
+
+    hash_type = request.args.get('hash_type', type=int)
+    if hash_type is not None:
+        return jsonify({'status': 200, 'hash_type': hash_type,
+                        'speed': slowest_benchmark(hash_type)})
```
i128: What happens if no results are returned?

i128:
please address the two comments and update the merge conflict.

hashview/api/routes.py
bandrel: Done in 4eaede2 — switched to `agent=False` and dropped the `update_heartbeat()` call. Agreed on both counts: an agent has no use for the fleet-wide slowest-per-type view, and stamping `last_checkin` from a plain read would misreport whoever called it as a live agent.

OpenAPI updated to match (`x-audience: user`, `userApiKey` only), plus a test that an agent uuid gets the 302 to `/v1/not_authorized` and one asserting `last_checkin` is untouched.

Note `GET /v1/tasks` in this same PR still uses `agent=True`, matching its siblings (`/v1/tasks/<id>`, `/v1/customers`). Happy to flip that to user-only too if you want it consistent — say the word.

hashview/api/routes.py
bandrel: Two empty cases, now handled differently on purpose:

- **`?hash_type=<mode>` with no benchmark** — was 200 with `speed: null`, forcing every caller to special-case a null. Now a 404 (`{status: 404, type: Error, msg: 'No benchmark recorded for hash type N'}`).
- **No query param, empty table** — stays 200 with `performance: {}`. Zero benchmarks isn't an error, and this matches the other collection endpoints returning empty lists.

Both are covered by tests (`test_get_benchmark_single_hash_and_missing`, `test_get_benchmark_empty_table`), and the spec's `speed` is no longer nullable.

bandrel:
Both inline comments addressed in 4eaede2: the benchmark endpoint was switched to `agent=False` with `update_heartbeat()` removed (agents have no use for the fleet-wide view, and stamping `last_checkin` from a read would misreport the caller as live), and empty-result cases are now handled explicitly — a `?hash_type=` with no recorded benchmark returns 404 instead of 200 with a null speed, while a bare collection request still returns 200 with an empty map.

The merge conflict (duplicate `regardless` entry in `.githooks/safe-allowlist.txt`) is resolved; head is at 4eaede2.

All CI checks pass (bandit, build 3.11/3.12/3.13, e2e, e2e-crack, mariadb-parity, migration-e2e, openapi, pip-audit, ruff, unit 3.11/3.12/3.13). coverage-badge is skipping as expected on non-main branches.

Ready for another look.

Issue #281, opened 2026-06-22
Agent process-control reliability: false self-kill on hashcat stderr + broken job cancellation


bandrel:
## Summary

Two distinct reliability bugs in the agent's process control. Both are independent of the e2e test work (PR #279) and live in `install/hashview-agent/hashview-agent.py`. Grouped because both concern how the agent runs and controls the hashcat subprocess.

---

## D. `run_command` treats ANY stderr output as fatal → false self-kill

`hashview-agent.py:118-134`:

```python
def run_command(command):
    try:
        cmd = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
        _output, error = cmd.communicate()
        if error:
            LOG.error('Command stderr: %s', error.decode('utf-8', 'replace').strip())
            if 'hashfile is empty or corrupt' not in str(error):
                if 'Terminated' in str(error):
                    sys.exit()
                else:
                    api.sendError(str(error))
                    os.kill(os.getpid(), signal.SIGINT)   # agent kills itself
```

Any bytes on stderr that aren't the two special-cased strings cause the agent to report an error and **SIGINT itself**. But real hashcat routinely writes **benign** messages to stderr — OpenCL/CUDA/driver notices, "Not enough allocatable device memory" warnings, deprecation notices, etc. A normal, successful run can therefore kill the agent mid-job.

(Decision to treat stderr as fatal is also why a drop-in test shim must stay completely silent on stderr — a sign the contract is too strict.)

**Suggested fix:** Don't treat non-empty stderr as failure. Key off the process **exit code** (`cmd.returncode`), and/or match known fatal error signatures, while logging other stderr as warnings. Never `os.kill(SIGINT)` on benign output.

---

## E. `getHashcatPid` aborts its scan on the first inaccessible process → cancellation silently fails

`hashview-agent.py:139-154`:

```python
        for proc in psutil.process_iter():
            try:
                pinfo = proc.as_dict(attrs=['pid', 'name', 'cmdline'])
                if 'hashcat' in pinfo['name'].lower():
                    for cli_args in pinfo['cmdline']:
                        if 'hc_cracked_' in cli_args:
                            return pinfo['pid']
            except:
                return False          # <-- aborts the WHOLE scan, inside the loop
    return False
```

[...]

PR #366, opened 2026-07-24
fix(agent): stop shelling out during rule sync; bound the server-supplied filename

bandrel:
## Problem

The agent's rules sync installed downloaded rule files by shelling out:

```python
run_command(f'gunzip {tmp_gz}')
run_command(f'mv {tmp_file} {dest}')   # dest built from the server's Rules.path
```

`run_command` was `subprocess.Popen(command, shell=True)`, so a value from the server's database reached a shell command line as the agent's user. Post-#337 servers randomize the stored path — but installs upgraded from before #337 still carry attacker-influenced rows, and current reachability is not a boundary. This fixes the mechanism rather than arguing the input is benign today.

## Change

- **Shell-free install:** `gzip.decompress` + `os.replace`. The two duplicated download blocks collapse into one `_install_rule_file` helper.
- **`run_command` deleted** — it was the last `shell=True` sink in the agent (the crack path already runs argv with `shell=False` via `run_hashcat`). Bandit on the agent is now down to the low-severity `B404` import notice.
- **`_safe_control_filename`:** the untrusted `path` from a rules *or* wordlists manifest entry is reduced to a plain filename; an entry that doesn't reduce to one (`''`, `.`, `..`, embedded NUL) is skipped with a warning instead of being written.
- **Robustness side benefit:** a corrupt/non-gzip download now skips that one rule. Previously `gunzip`'s stderr hit `run_command`'s fatal-stderr branch and SIGINT'd the whole agent.
- **`Dockerfile.agent`** drops its `gzip`/`coreutils` layer — the agent needs no external binaries for sync.

## Tests

`tests/agent_unit/test_sync_no_shell.py` (15 cases):

- `Popen`/`os.system` are hard failures for the duration of the sync
- a metacharacter-laden rule *and* wordlist name lands as one literal file inside `control/`, with no side effects
- non-basename paths are refused; `../../../../etc/passwd` collapses to `passwd` inside `control/rules`
- checksum mismatch and corrupt gzip both discard cleanly, leaving `control/tmp` empty
- no call anywhere in the agent passes a truthy `shell=` (checked via the AST, so prose about the old behaviour is fine)

Mutation-verified: dropping the sanitizer's reject list fails 5 of them.

Also repoints the `#281 D` strict xfail at `run_hashcat`. It called the now-deleted `run_command`, and an `AttributeError` still counts as an expected failure — so it would have kept reporting `xfailed` while exercising nothing. Confirmed it still fails for the real reason (`killed == [(pid, SIGINT)]`).

## Verification

`tests/unit tests/security tests/agent_unit` — 1443 passed, 2 skipped, 58 xfailed, 5 xpassed (xpasses pre-existing). `tests/unit/test_wordlist_gzip_storage.py`'s AST extractor needed the new helper added to its `wanted` set.


i128:
Good find/fix. thanks

PR #367, opened 2026-07-25
fix(api): remove control/tmp scratch files after download responses

### LINKED ISSUE #226: API: temp files in control/tmp are never cleaned up (disk leak)
## Summary
Rule download (`hashview/api/routes.py:409`), wordlist download (`:525`), and hashfile generation (`:1124`) write files into `control/tmp/` and never delete them. This causes steady disk growth.

## Suggested fix
Delete the temp file after sending, or implement a periodic cleanup of `control/tmp/` (the scheduler already runs a data-retention job).

Sanjays2402:
Closes #226

The rules, wordlist, and hashfile download endpoints each write a file into `control/tmp` purely to stream it back, but never delete it, so the directory grows on every agent poll. Each one now registers an `after_this_request` hook that removes its own temp file once the response has been sent.

New `tests/unit/test_api_tmp_file_cleanup.py` asserts the `control/tmp` listing is unchanged across each of the three downloads; all three fail on current `main` and pass with the fix (`tests/unit`: 16 -> 19 passed).

### DIFF
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -1,4 +1,4 @@
-from flask import Blueprint, jsonify, redirect, request, send_from_directory, current_app, url_for
+from flask import Blueprint, jsonify, redirect, request, send_from_directory, current_app, url_for, after_this_request
 from hashview.models import Agents, JobTasks, Tasks, Wordlists, Rules, Jobs, Hashes, Hashfiles, HashfileHashes, Users, HashNotifications, Settings, JobNotifications, Customers
 from hashview.utils.utils import get_md5_hash, get_linecount, get_filehash, update_dynamic_wordlist, update_job_task_status, import_hashfilehashes, validate_pwdump_hashfile, validate_netntlm_hashfile, validate_kerberos_hashfile, validate_shadow_hashfile, validate_user_hash_hashfile, validate_hash_only_hashfile, send_email, send_pushover, notify_admins, build_hashcat_command, ntlm_hash_hex
 from hashview.models import db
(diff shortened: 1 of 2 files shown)


Yoshi325:
@Sanjays2402 please re-point this to the current version branch. `main` is our release branch, and we develop on the latest version branch.

bandrel:
Heads-up: this fix already landed on `v0.8.3-dev` by another route, so this PR is now superseded. `_remove_file` / `_send_generated_file` (`hashview/api/routes.py:174-189`) register a `response.call_on_close` cleanup hook, and all three download sites use it — rules `:703`, wordlists `:831`, hashfile generation `:1465` — with the founds-import path cleaning up in a `finally` at `:1980`. I have closed #226 citing that code.

Recommend closing this without merging unless it covers a site the merged version missed.

bandrel:
Closing as superseded — the fix landed on `v0.8.3-dev` by another route before this could merge.

This PR adds `remove_after_request()` and wires it into three sites: rules download, wordlist download, and hashfile generation. The branch now does the same job with `_remove_file` / `_send_generated_file` (`hashview/api/routes.py:174-189`), which registers a `response.call_on_close` hook, used at exactly those three sites — rules `:703`, wordlists `:831`, hashfile `:1465` — plus the founds-import path cleaning up in a `finally` at `:1980`. I checked this PR's diff site-for-site against the merged version and there is no site here that the branch misses. #226 is closed citing that code.

**One thing here is not redundant and shouldn't be lost with it.** The merged fix shipped without a regression test — there is no test anywhere in `tests/unit` or `tests/security` asserting the `control/tmp` listing is unchanged across a download, so someone could delete the `call_on_close` hook tomorrow and CI would stay green. The three tests in `tests/unit/test_api_tmp_file_cleanup.py` on this branch are still worth having, rebased onto the current helper names. Worth cherry-picking into a small test-only PR rather than discarding.

bandrel:
Correction to my earlier comment on this PR: I closed it as superseded, and that was wrong. Your `after_this_request` approach was the correct one; the `call_on_close` version I closed it in favour of never actually runs.

`send_from_directory` sets `direct_passthrough=True`, and werkzeug's `Response.get_app_iter` returns the raw file wrapper for that case instead of wrapping it in a `ClosingIterator` (`werkzeug/wrappers/response.py:541-545`), so `Response.close()` — and every `call_on_close` callback with it — is never invoked by the WSGI layer. The temp files were still accumulating.

I have reopened #226 and opened PR #420, which uses `after_this_request` exactly as you did here, plus your three tests (they needed the test-client cookie domain to authenticate, and the wordlist case needed a dynamic wordlist to actually hit the temp-file path). Credit for the mechanism is yours.

PR #380, opened 2026-08-13
Add tests documenting lucky-task assignment count bug (#379)

### LINKED ISSUE #379: "I'm Feeling Lucky — top 5" button actually adds top 10 tasks
**Summary**
The "I'm Feeling Lucky — top 5" button on the job task-assignment page adds 10 tasks instead of 5.

**Location**
- Button label: `hashview/templates/jobs_assigned_tasks.html.j2:73`
- Backend handler: `hashview/jobs/routes.py:598` (`jobs_assign_lucky_task_group`)

**Details**
The query in the handler is hard-coded to `.limit(10)` (`routes.py:620`), and the success flash message even confirms it: `'Successfully Added Top 10 Tasks'` (`routes.py:634`). The button text ("top 5") was never updated to match, so the UI promises 5 tasks but the backend assigns up to 10.

**Steps to reproduce**
1. Open a job's task queue page.
2. Click "I'm Feeling Lucky — top 5".
3. Observe up to 10 tasks get added (and the flash message says "Top 10").

**Expected behavior**
Either:
- (a) the button/UI should say "top 10" to match current backend behavior, or
- (b) the backend `.limit(10)` should be changed to `.limit(5)` to match the button label.

**Suggested fix**
Whichever count is intended, align the button label, backend `.limit()` call, and flash message so all three agree.

bandrel:
## Summary
- Fixes #379: the "I'm Feeling Lucky" button labeled itself "top 5" while the backend (`jobs_assign_lucky_task_group`, hashview/jobs/routes.py) has always queried and assigned the top 10 historically effective tasks (or fewer, if fewer exist). The backend behavior was correct; the label was the bug. Updated `jobs_assigned_tasks.html.j2` to say "top 10".
- `test_jobs_tasks_page_lucky_button_label_matches_backend_top_ten` (new, `tests/unit/test_jobs_routes.py`): asserts the rendered button text says "top 10" and not "top 5".
- `test_lucky_assigns_all_when_fewer_than_ten_effective_tasks_exist` (`tests/unit/test_issue_xfail_misc.py`): replaces the earlier `test_lucky_task_assignment_adds_at_most_five_tasks` xfail (which assumed the fix would cap the backend at 5). Seeds 7 effective tasks and asserts all 7 are assigned, locking in the "fewer than 10 available -> assign all of them" behavior.
- `test_lucky_caps_at_ten_tasks_when_more_are_effective` (`tests/unit/test_lucky_and_one_and_done.py`, pre-existing): seeds 12 effective tasks and asserts only the top 10 are assigned. Docstring updated to drop the stale "future fix" framing now that the fix has landed.
- CHANGELOG entry added under the unreleased v0.8.3-Beta "Fixed" section.

## Test plan
- [x] `pytest tests/unit/test_jobs_routes.py tests/unit/test_lucky_and_one_and_done.py tests/unit/test_issue_xfail_misc.py -v` — 30 passed, 2 xfailed (unrelated pre-existing), 2 xpassed (unrelated pre-existing)
- [x] Full `tests/unit tests/security tests/agent_unit` suite passes via pre-push hook


Yoshi325:
We should consider `isort` or something like to to make sure that import statements stay ordered and formatted correctly. This has a little bit of churn in it due to import re-write. Not a block, but something to note.

Yoshi325:
Two new tests make different claims about the same endpoint. The fix for issue #379 will break one of them.

The test `test_lucky_task_assignment_adds_at_most_five_tasks` asserts that the endpoint assigns 5 tasks. Issue #379 asks for this behavior. The xfail mark on this test is correct. The second new test is in `test_lucky_and_one_and_done.py`. It asserts that the endpoint assigns exactly 10 tasks, and that the assigned ids equal the top 10 ids. These two tests cannot both be correct.

To check, I changed line 620 of `hashview/jobs/routes.py` from `.limit(10)` to `.limit(5)`. Then I ran both tests again. The xfail test became an XPASS. The cap test failed with `assert 5 == 10`.

The docstring makes this a blocker. It says that the test prevents an unbounded query after a future fix. The goal is correct, but the assertion is wrong. A fixed count of 10 does not detect an unbounded query. It detects all changes, and this includes the correct change.

Use a range instead. Assert that the endpoint assigns 5 to 10 tasks, and that all assigned ids come from the top 10. Assert also that no id comes from a lower rank. These assertions detect a query that returns all 12 tasks. They stay true if the limit becomes 5, and also if it stays at 10.

The second blocker is a test that is not a focus of this PR. Line 750 of `tests/unit/test_jobs_routes_branches.py` asserts the exact bytes `Successfully Added Top 10 Tasks`. This flash message is not a small detail of issue #379. It is one half of the bug, because the route shows 10 and the button shows 5. A correct fix must change this message, and the change will break the assertion.

In summary, we should do two things to get this PR ready to merge. First, write the cap test with a range instead of a constant. Second, make the flash assertion independent of the count, or list in the PR description each test that the fix must change.

Yoshi325:
Three items remain before this PR is ready to merge.

First, the fix has no test. The three lucky tests all check the backend count. That count was correct before this PR, and it does not change here. The only production change is line 73 of `hashview/templates/jobs_assigned_tasks.html.j2`. No test asserts the new label. The command `git grep "Feeling Lucky" -- tests/` returns one comment and no assertion.

The label and the backend disagreed from commit `3c0b871` in June 2026 until now. No test caught that error. Add a plain assertion for the label. Do not use an xfail, because the fix is in this PR. The file `tests/unit/test_jobs_routes.py:238` shows the pattern:

```python
html = client.get(f"/jobs/{job.id}/tasks").get_data(as_text=True)
assert "top 10" in html
```

Second, the docstring of test_lucky_caps_at_ten_tasks_when_more_are_effective is out of date. Lines 154 to 156 refer to a future fix for issue #379, and they say that the button shows "top 5". This PR is that fix. The button now shows "top 10". Please update the text.

Third, the PR description is out of date in three places. It describes test_lucky_task_assignment_adds_at_most_five_tasks, but this PR deletes that test. It says "No production code changes", but this PR changes a template and the CHANGELOG. The test plan reports 1 xfail, but the new test is not an xfail.

bandrel:
@Yoshi325 addressed the 18:02 UTC comment (all three items pushed in 7ff206c):

1. Added `test_jobs_tasks_page_lucky_button_label_matches_backend_top_ten` in `tests/unit/test_jobs_routes.py` — a plain (non-xfail) assertion on the rendered `/jobs/<id>/tasks` HTML, following the `body.count(...)`/`assert "..." in body` pattern already used in that file (e.g. `test_jobs_list_collapses_chunked_task_to_one_card`). Confirmed with `git grep "Feeling Lucky" -- tests/` that no test previously asserted the label text.
2. Updated the docstring on `test_lucky_caps_at_ten_tasks_when_more_are_effective` — it referred to "a future fix for issue #379 (button says 'top 5')"; that fix is this PR, so the button now says "top 10". Reworded to point at the new label test instead.
3. Rewrote the PR description: dropped the reference to the deleted `test_lucky_task_assignment_adds_at_most_five_tasks`, corrected "No production code changes" (the template and CHANGELOG do change), and fixed the test-plan counts.

On the 16:09 UTC comment (range-vs-exact-count on `test_lucky_caps_at_ten_tasks_when_more_are_effective`): that was against the pre-fix state, where "10" might have been a placeholder a #379 fix would shrink to 5. The resolution here keeps the backend's `.limit(10)` and flash message unchanged — the label was the actual bug, not the count — so the exact-count assertion is correct as the final invariant, not something a legitimate future fix needs to loosen. Left as-is.

isort noted, treating as out of scope for this fix.

i128:
Awesome work. Thanks!

Issue #385, opened 2026-08-14
Analytics double-counts every plaintext-derived figure when a hash belongs to multiple hashfiles (reuse donut reads 100%)


bandrel:
## Summary

Every plaintext-derived figure on `/analytics` is computed from one row **per `hashfile_hashes` join row**, not per hash. The `Hashes -> HashfileHashes` join is one-to-many, so a hash that belongs to more than one hashfile is counted once per hashfile it appears in. On a dataset where each hash sits in two hashfiles, every one of those figures is inflated 2x — and the Password Reuse donut reads **100% reused / 0% unique**, because each hash "shares" a password with its own duplicate row.

This is the same root shape as the shared-password miscount fixed on `fix/analytics-page-hang`, but that fix only corrected the Shared Passwords card. The rest of the page still counts join rows.

## The corpus

`hashview/analytics/routes.py:328-331`:

```python
corpus = [(decode_hex_plain(plaintext), username)
          for plaintext, username in
          _scoped_hash_query(customer_id, hashfile_id, cracked=True)
          .with_entities(Hashes.plaintext, HashfileHashes.username).all()]
total_cracked = len(corpus)
```

There is no `.distinct()` here, so `len(corpus)` is a count of join rows. Compare `_distinct_hash_count()` (line 70), which *does* dedupe on `Hashes.id` and is used for the summary tiles — so the page already disagrees with itself.

## Measured on a 34,450-hash dataset

Each hash in this dataset belongs to 2 hashfiles:

```
total_cracked (len(corpus))   = 68898
unique_cracked (distinct id)  = 34449     <- the real number
accounts (join rows)          = 68900
distinct plaintexts           = 34449
top password freq             = every top entry has a count of 2
reused_pct                    = 100.0
```

Every plaintext in the dataset is distinct, yet the page reports **100% password reuse and 0% unique**, and each "top recovered password" shows a count of 2. The reuse donut is the most visibly wrong, but it is not a special case — it is just the figure where the doubling is most obvious.

## Affected figures

All of these consume `corpus` (or `total_cracked`) and are inflated:

| Figure | Source |
|---|---|
| Password Reuse donut (`reused_pct` / `unique_pct` / `top_reuse_count`) | `routes.py:363-366` |
| Top Recovered Passwords | `freq`, `routes.py:362` |
| Password Length Distribution | `length_counter`, `routes.py:346` |
| Top Masks | `mask_counter`, `routes.py:347` |
[...]

PR #391, opened 2026-08-15
fix(analytics): stop /analytics hanging the browser (50MB response) and miscounting shared passwords

bandrel:
## The hang

On a production-sized hashfile, `/analytics` locked the browser up. The server wasn't the problem — it produced the page in 1.6s. The response was **50.2MB and roughly 350k DOM nodes**.

The Shared Passwords and Username = Password cards are the only lists on the page whose length scales with the dataset (everything else is a fixed number of buckets), and each shared row is a whole `<form>` with three hidden inputs. Against a 34,449-hash dataset the page rendered 34,449 forms.

Two separate defects produced that number, so this is two commits.

### 1. Shared passwords were counted per join row, not per account

`Hashes -> HashfileHashes` is one-to-many, so a hash belonging to two hashfiles yields two rows for the *same* account. `shared_map` collected those in a list, so every password on a two-hashfile hash looked like it was "shared" — by one account with itself. Rows with no username at all were counted too, even though an unnamed row identifies no account and can't establish sharing.

`shared_map` and `_shared_groups()` (the zip download) now collect usernames in a **set** and skip unnamed rows, so the card and the download agree and a single account can't share a password with itself.

This is a slice of the same root cause as #385, which covers the rest of the page — the counters this PR doesn't touch still count join rows.

### 2. The cards rendered every row

Both cards now render at most `PREVIEW_LIMIT` (100) rows, report the **true** total in the header, and say how many are shown. The full sets are unchanged and still available from the existing `shared_zip` / `fig8` download buttons.

Same dataset after both fixes: **0.14MB**.

## Verification

- `pytest tests/unit tests/security tests/agent_unit` — 1518 passed, 2 skipped, 39 xfailed, 5 xpassed (the xpasses are pre-existing on `v0.8.3-dev`)
- `ruff check hashview/ hashview.py` clean; `pylint` 9.63/10, unchanged
- 99 lines of new unit coverage in `tests/unit/test_analytics_view.py` pinning both behaviours: a shared group counts distinct named accounts, unnamed rows never form a group, and each card renders at most `PREVIEW_LIMIT` rows while reporting the full total
- Built a docker image from this branch and ran the `/analytics` docker suite from #390 against it: the `test_shared_card_caps_rendered_rows_but_reports_the_true_total` strict xfail flips to **XPASS**, which is the signal that the fix works

## Stack

#390 (the `/analytics` docker bug suite) is stacked **on top of** this branch, and drops that test's xfail because the cap lands here. Merge this one first; GitHub will retarget #390 to `v0.8.3-dev` automatically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,8 @@ Notable changes will be documented here
 - Pinned Python to 3.11+; added a ruff / pylint / bandit / pre-commit lint-and-security stack, a LICENSE, and substantially expanded automated tests (unit, agent, end-to-end, and security)
 
 ### Fixed
+- The Analytics page no longer hangs the browser on large datasets: the Shared Passwords and Username = Password cards render a capped preview (with the full total shown) instead of one form per group, and the complete lists remain available from the download buttons
+- Analytics counted a password as "shared" when a single account's hash appeared in more than one hashfile, or when the rows had no username at all; shared passwords are now grouped by distinct named account
 - The agent's HTTP/API layer is now resilient to non-200 and unexpected server responses instead of crashing with opaque errors
 - Command-injection hardening: hashcat is invoked as an argv list with no shell
 - CSRF: state-changing actions were moved from GET to POST and protected with CSRF tokens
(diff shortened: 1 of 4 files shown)


i128:
Good find/fix.

Issue #377, opened 2026-07-30
Website Keywords: a crawl that raises serves the previous run's stale wordlist


bandrel:
## Summary

The `(DYNAMIC) Website Keywords` wordlist is meant to be regenerated from scratch on every agent request — crawl the job's `crawl_url`, write the result to a randomly-named file under `hashview/control/tmp/`, then atomically `os.replace` it onto `wordlist.path`. An agent should never receive a cached or stale list.

That holds when the crawl *returns nothing*, but not when the crawl *raises*.

## Details

In `_generate_website_keywords` (`hashview/utils/utils.py:703`):

```python
words = crawl_website_keywords(target, settings)          # line 722 — unguarded

tmp_path = os.path.join(current_app.root_path, 'control/tmp', secrets.token_hex(8) + '.txt')
with open(tmp_path, 'w') as tmp:
    for word in sorted(words):
        tmp.write(word + '\n')
try:
    os.replace(tmp_path, wordlist.path)
except OSError:
    shutil.move(tmp_path, wordlist.path)
```

If `crawl_website_keywords` raises, the exception propagates out of `_generate_website_keywords` and out of `update_dynamic_wordlist`, so:

1. The tmp file is never written and `os.replace` never runs — `wordlist.path` still holds **the previous job's crawl output**.
2. The metadata refresh at `hashview/utils/utils.py:788-798` (`size` / `checksum` / `byte_size` / `last_updated`) never runs.
3. `GET /v1/wordlists/<id>` gzips and serves that stale file to the agent, which cracks against another engagement's keywords.

`crawler.py` swallows per-page `requests` errors inside `_fetch_and_parse`, so this is not the common "site is down" case (that correctly yields an empty set and a blank list). It is reachable via anything that escapes the per-page guard — a `ThreadPoolExecutor` failure, `MemoryError` on a huge response, or an `OSError` creating/writing the tmp file itself.

## Expected

A failed crawl must produce a **blank** wordlist, with the `Wordlists` row metadata updated to describe that blank file, and no tmp debris left in `control/tmp/`.

## Suggested fix

Wrap the crawl so a failure degrades to an empty result rather than aborting the regeneration, and clean up the tmp file on any error:

- `try: words = crawl_website_keywords(...) / except Exception: log + words = set()`
- write/replace inside `try ... finally`, unlinking `tmp_path` if it still exists

## Tests

Strict-`xfail` regression tests are in `tests/unit/test_website_keywords_no_stale.py`:

[...]

PR #400, opened 2026-08-19
Fix website keywords job collision

Yoshi325:
## The Bug

`(DYNAMIC) Website Keywords` is one `Wordlists` row with one file, but its
content comes from each job's `Jobs.crawl_url`. Concurrent jobs overwrite each
other: job A crawls `acme.example`, job B replaces the file with
`othercorp.example`, and job A's agent then cracks against othercorp's
keywords.

Nothing reports an error. The job completes and just finds fewer hashes. On an
engagement it puts one client's crawled keywords onto a job for another.

## The Fix

Each job gets its own copy, resolved per caller. No schema change, no
migration, no agent change.

- `is_job_scoped_wordlist()` + `job_wordlist_path()` materialize
  `dynamic-website-keywords-job<id>.txt`. Only Website Keywords is job-scoped;
  the DB-derived lists (Passwords, Usernames, Customers, NTLM) stay global.
- `GET /v1/wordlists` reports the checksum of the caller's copy, so the agent's
  cache invalidates when it moves between jobs.
- `GET /v1/wordlists/<id>` serves the caller's copy, and 404s rather than
  falling back to another job's file.
- All four teardown paths (UI delete, API delete, both retention sweeps) call
  `remove_job_wordlist_files()`.

The second commit collapses two different queries for "which job is this agent
on" (heartbeat re-issue vs `/v1/updateWordlist`) onto one helper, so they can
no longer disagree.

## The Support

1462 passed, 2 skipped, 36 xfailed, 5 xpassed.

New: cross-job isolation end to end through the real endpoints, job deletion
over both UI and API, and heartbeat/wordlist job agreement.

## Other Bits and Bobs

- The row's `path` is unchanged on purpose. The agent's local filename and the
  basename in `build_hashcat_command` both derive from it, so only the bytes
  differ per job.
- Row `size`/`checksum`/`byte_size` now describe the last file generated. A UI
  download of Website Keywords serves the row file, which is no longer written.
- `test_website_keywords_no_stale.py` (#378) asserted against the row path.
  Those assertions move to the job's copy; the contract is unchanged and the
  #377 xfails still xfail.
- The two `scheduler.py` retention sweeps call the same cleanup helper as the
  tested delete routes but have no unit coverage of their own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Notable changes will be documented here
 - Pinned Python to 3.11+; added a ruff / pylint / bandit / pre-commit lint-and-security stack, a LICENSE, and substantially expanded automated tests (unit, agent, end-to-end, and security)
 
 ### Fixed
+- Two jobs using the (DYNAMIC) Website Keywords wordlist no longer overwrite each other's crawl results. Each job now gets its own copy of the list, so an agent is always served the keywords crawled for the job it is actually running; previously whichever job regenerated last won, and the other job cracked against the wrong site's keywords without reporting an error. A job's copy is crawled once and removed when the job is deleted or ages out under the retention policy.
 - The agent's HTTP/API layer is now resilient to non-200 and unexpected server responses instead of crashing with opaque errors
 - Command-injection hardening: hashcat is invoked as an argv list with no shell
 - CSRF: state-changing actions were moved from GET to POST and protected with CSRF tokens
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -225,6 +225,23 @@ def v1_api_get_admin_settings():
 _ACTIVE_JOBTASK_STATUSES = ('Running', 'Queued', 'Not Started', 'Importing')
 
 
+def _agent_active_job_task(agent_id):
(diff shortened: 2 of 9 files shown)


bandrel:
Flagging a mismatch between this PR's body and its diff, so #377 doesn't get auto-closed on merge.

The body references #377 ("a crawl that raises serves the previous run's stale wordlist"), but the diff doesn't address it. `words = crawl_website_keywords(target, settings)` is still an unguarded call — it appears in the diff only as context, not as an added line — and the `try`/`except` blocks this PR does add are `except OSError` around the tmp-file write and move, not around the crawl. So if the crawl raises, the exception still propagates and the previous run's file is still what agents get.

What this PR actually fixes is the job-collision bug (one shared `(DYNAMIC) Website Keywords` row and file, concurrent jobs overwriting each other's crawl), which is real and worth merging on its own. Just worth merging it as fixing the collision and leaving #377 open, rather than closing both.

i128:
So i think based on our internal conversations, we're thinking we just strip web scraping as a dyanmic list option. So if thats correct we can close this PR.

Issue #22, opened 2022-02-03
User search is case sensitive, should be case insensitive


i128:
specific issue with with hashview/hashview/searches/routes.py

```
if searchForm.validate_on_submit():
        if searchForm.search_type.data == 'hash':
            results = db.session.query(Hashes, HashfileHashes).join(HashfileHashes, Hashes.id==HashfileHashes.hash_id).filter(Hashes.ciphertext==searchForm.query.data).all()
        elif searchForm.search_type.data == 'user':
            results = db.session.query(Hashes, HashfileHashes).join(HashfileHashes, Hashes.id==HashfileHashes.hash_id).filter(HashfileHashes.username.like('%' + searchForm.query.data.encode('latin-1').hex() + '%')).all()
        elif searchForm.search_type.data == 'password':
            results = db.session.query(Hashes, HashfileHashes).join(HashfileHashes, Hashes.id==HashfileHashes.hash_id).filter(Hashes.plaintext == searchForm.query.data.encode('latin-1').hex()).all()
```

This is because we hex encode the string prior to doing a search. Unfortunately to really fix this we'd either need to hex decode every entry in hashfile_hashes.username and then compare.

another option might be to force lowercase usernames on import and then force lower case searching 🤔

bandrel:
Verified fixed in v0.8.3-dev: user search now matches case-insensitively (latin-1 hex-encoding replaced by a UTF-8 path; SQLite LIKE is case-insensitive for ASCII). Added a regression guard (xpass) in tests/unit/test_issue_xfail_search_encoding.py::test_user_search_is_case_insensitive.

Issue #364, opened 2026-07-24
Move hashfile import to the background so users don't wait on large hashsets


bandrel:
## Summary
Hashfile import runs synchronously inside the upload POST (`hashview/jobs/routes.py:256`), so the user's browser blocks for the entire import. On a large hashset this is minutes-to-hours. Move the heavy import work to the background after the job/hashfile is created so the user does not wait.

## Existing precedent to reuse (no new deps)
The codebase already has a proven async pattern in **wordlist import**: validate quickly, spawn a daemon thread, return immediately (`hashview/wordlists/routes.py:124-128`), with `run_import_async` opening a fresh `app.app_context()` for a clean scoped DB session (`hashview/utils/wordlist_import.py:159`).

## Proposed approach — mirror that pattern for hashfile import
1. **Keep validation synchronous** so a bad file / hash-type fails instantly in the request; only the heavy per-line import + dedup moves to the thread.
2. **Add hashfile import status** — `Hashfiles.import_status` (`importing` -> `ready` / `failed`) plus `imported_count` / `total_count`; a small poll endpoint can feed a progress UI its real percentage.
3. **Gate job dispatch on `ready`** — critical correctness item: agents must not be handed chunks for a hashfile still importing. Add the `import_status == 'ready'` check wherever a job is deemed runnable.
4. **Background error handling** — on mid-import failure set `import_status='failed'` + reason (mirrors the wordlist path's `.fail` rename + summary).

## Trade-off / known limitation
The daemon-thread approach (consistent with the existing wordlist convention) is **not durable**: a web-process restart mid-import strands the hashfile in `importing`. Mitigation: a startup sweep that fails/re-queues stranded `importing` rows. Full durability (survives restarts, retries) would require a real task queue (Celery/RQ + broker) — a much larger infra change, out of scope here.

## Dependency
Should land on top of the batched-commit optimization (#363) so the background job finishes in minutes and holds a narrow transaction.

bandrel:
Cross-referencing a test-cleanup decision that touches this issue's item 2 (`Hashfiles.import_status`).

[...]

Issue #176, opened 2026-06-04
Add status indicator for large hashfile imports


i128:

i128:
Done

Issue #140, opened 2024-12-10
Usernames with the en dash u2013 can not be imported


i128:
This is due to hashview trying to latin-1 encode the character which does not exit. Maybe switch to a different char encoding?

bandrel:
Verified fixed in v0.8.3-dev: username import no longer latin-1 encodes; normalization goes through text_from_field/bytes_to_text (UTF-8 + $HEX[...] fallback), so en dash (U+2013) round-trips. Added a regression guard (xpass) in tests/unit/test_issue_xfail_search_encoding.py::test_username_with_en_dash_round_trips.

PR #406, opened 2026-08-20
test: drop stale xfail markers for closed issues #22, #99, #100, #140

bandrel:
## Summary

Cleans up xfail markers that no longer describe reality.

**Converted to plain regression guards** — four tests have been XPASSing since their issues were fixed and closed (#22, #99, #100, #140). Their non-strict `xfail` markers meant a future regression would silently XFAIL instead of failing CI:

- `test_task_group_can_be_renamed_after_creation` (#100)
- `test_job_task_selection_lists_all_tasks` (#99)
- `test_user_search_is_case_insensitive` (#22)
- `test_username_with_en_dash_round_trips` (#140)

**Deleted** — `test_hashfile_exposes_import_status` (#176). That issue was closed as completed and the indicator did ship, as the stepped "Importing hashfile" progress modal in `jobs_assigned_hashfiles.html.j2` (~265-393) driven by `xhr.upload.onprogress`. The test asserted a persisted `Hashfiles` column instead, which was never the delivered design, so it XFAILed against a premise that no longer holds. A persisted import status is only needed for background imports (#364) and belongs there.

Docstrings and module headers updated to match; the `pytest` import dropped from the search-encoding file where no markers remain. The #57 xfail is left in place — that issue is still open and still reproduces.

## Testing

- Both files: 6 tests → 5 passed, 1 xfailed (the intentional #57)
- Full pre-push gate on each commit: ruff + bandit + openapi + 1518 passed, 38 xfailed

### DIFF
--- a/tests/unit/test_issue_xfail_misc.py
+++ b/tests/unit/test_issue_xfail_misc.py
@@ -95,6 +95,34 @@ def test_db_password_with_percent_is_parsed():
 
 
 # ---------------------------------------------------------------------------
+# Issue #364 — "Move hashfile import to the background"
+# ---------------------------------------------------------------------------
+@pytest.mark.xfail(reason="issue #364: no persisted hashfile import status",
+                   strict=True)
+def test_hashfile_exposes_persisted_import_status(app):
+    """The Hashfiles model should persist the state of a running import.
+
+    Import runs synchronously inside the upload POST today, so progress is
+    reported only by the client-side modal (guarded in
+    test_jobs_assigned_hashfile.py). Once #364 moves the work to a
+    background thread the response returns first and that modal can no
+    longer observe the import — item 2 of #364 adds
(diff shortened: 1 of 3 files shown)


bandrel:
Follow-up commit 61f9e10 replaces the deleted #176 xfail with the two tests it should have been:

- **`test_page_renders_the_import_progress_modal`** — pins the shipped #176 feature: the `hf-import-modal` dialog and its `hf-step-upload`/`hf-step-import` steps on `GET /jobs/<id>/assigned_hashfile/`. Mutation-verified (renaming the modal id in the template fails the test).
- **`test_hashfile_exposes_persisted_import_status`** — a **strict** xfail reserving `Hashfiles.import_status` from #364 item 2, so the moment that column lands the test fails loudly and gets rewritten into a real assertion instead of rotting as a stale XPASS like its predecessor.

Full gate re-run on push: 1519 passed, 39 xfailed.

Issue #402, opened 2026-08-19
Mask files bypass chunking and the ETA: planner sees keyspace 1, so a file of cheap masks runs whole on one agent


bandrel:
## Summary

The chunk planner and the runtime estimate both treat `Tasks.hc_mask` as a single mask string. When it holds a **mask file** path instead — which hashcat accepts in the mask argv position for `-a 3/6/7`, and which is exactly what #384 is about making possible — the planner computes a keyspace of **1** and never chunks the task, and the only ETA shown to the user reflects one line of the file rather than the whole file.

A mask file full of individually-cheap masks is the worst case: 1,000 masks that each take 30s on the slowest agent is 8+ hours of work that ships to a single agent as one un-chunked chunk, with a dashboard ETA that reads a few minutes and keeps resetting. With `max_runtime_tasks` set it never finishes — it is cancelled mid-file at the cap, having covered an unknowable prefix of the lines (see "What the runtime cap actually does here" below).

This is a follow-on to #384, not a duplicate of it. #384 explicitly scopes tasks out ("Tasks stay as they are: `hc_mask` holds `control/masks/<file>`. No `Tasks` migration."), which is the right call for that PR — but it means the store landing on its own turns a today-nobody-can-do-this into a silently-mis-scheduled attack, because the files will finally be on the agents.

## Why the keyspace is 1

`parse_mask` finds no `?` tokens in a path, so every character becomes a `('lit', c)` token, and `mask_keyspace` returns the empty product of charset sizes — `1` (`hashview/utils/chunking.py:117-126`). In the mask branch of `plan_chunks`, `total = 1` is trivially `<= target_candidates`, so it returns `[{}]`, meaning "run whole" (`hashview/utils/chunking.py:224-240`).

Verified directly against the module (it is dependency-free, so this runs standalone):

```
'control/masks/rockyou-1-60.hcmask' keyspace= 1  chunks= 1
'8char-1l-1d-1s.hcmask'             keyspace= 1  chunks= 1
'?d?d?d?d?d?d?d?d'                  keyspace= 100000000  chunks= 1   # correct: fits in one chunk
```

Note this is not a case the safety rails catch — it isn't "refuse to split something unsplittable" (which `parse_mask` returning `None` for `?b` / `?1-?4` handles deliberately and correctly). The planner believes it has a one-candidate task and that the answer is genuinely "no chunking needed."

[...]

PR #407, opened 2026-08-20
Add xfail coverage for the agent ETA clamp bug (#402)

bandrel:
Test-only. Adds `tests/agent_unit/test_issue_xfail_eta_clamp.py` pinning the agent-side ETA symptom of #402. No fix included — the xfails are the executable spec.

Refs #402.

## The bug

hashcat's `estimated_stop` is the absolute epoch at which the **current attack** ends. With a mask file every line is its own attack, so on the terminal status line of each mask that value equals the moment the mask finished — and the agent reads it up to one 15s poll later.

`time_difference()` (`install/hashview-agent/hashview-agent.py:533-541`) returns the sentence `"The specified time is in the past."` for any delta <= 0. `hashcatParser` wraps it in parens as `Time_Estimated` (`:598`), and the dashboard's `_eta_text` (`hashview/main/routes.py:170`) prints whatever is between the parentheses. The operator sees:

```
eta (The specified time is in the past.)
```

## This is ours, not hashcat's

At the identical boundary hashcat's own status output reads:

```
Time.Estimated...: Thu Aug 20 12:38:44 2026 (0 secs; Runtime limited: 30 secs)
Guess.Queue......: 1/3 (33.33%)
Progress.........: 100000000000/100000000000 (100.00%)
```

hashcat clamps the relative figure at zero. We reimplement the formatting and drop the clamp.

## Why it is not an edge case

Measured against hashcat `v7.1.2-484-g64e1bff93` (single GPU, `-m 0`, three long masks, `--status-timer=2`), each status line stamped with the wall-clock time it was emitted:

```
   t mask                        queue   prog%   est_stop - emitted_at
   0s ?d?d?d?d?d?d?d?d?d?d?d      1/3     9.8%        +9s
   3s ?d?d?d?d?d?d?d?d?d?d?d      1/3    32.0%        +6s
   5s ?d?d?d?d?d?d?d?d?d?d?d      1/3    54.1%        +3s
   7s ?d?d?d?d?d?d?d?d?d?d?d      1/3    76.2%        +2s
   9s ?d?d?d?d?d?d?d?d?d?d?d      1/3    95.9%         0s
  10s ?d?d?d?d?d?d?d?d?d?d?d      1/3   100.0%        -1s   <-- error string
  17s ?l?l?l?l?l?l?l              2/3    17.5%         0s
```

The countdown is accurate *while* a mask runs; only the terminal line goes negative. So this is a sawtooth that repeats once per line of the file — a file of long masks hits it as often as a file of cheap ones, just more slowly. Note also the 7-second hole between `t=10s` and `t=17s` while hashcat autotunes the next mask: no status line is written, and `hashcatParser` takes the last line in the file, so the terminal value is all that is available across that window.

## What the tests assert

| Test | Kind |
|---|---|
| `time_difference` on a completed attack (`== now`, 1s stale, 15s stale) | 3 × `xfail(strict=True)` |
| `hashcatParser` on a real captured terminal status line | `xfail(strict=True)` |
| A future `estimated_stop` still renders as a duration | passing regression guard |
| `guess_base_offset`/`count`/`percent` are present in the JSON | passing, documents available data |

`strict=True` matches the convention in `tests/agent_unit/test_issue_xfail_agent_robustness.py`: when the bug is fixed the tests pass and strict-xfail turns the unexpected pass into a failure, as a reminder to drop the markers.

The fixture is a real `--status-json` line captured from the run above, not hand-written. The target in it is hashcat's own documented example hash for `-m 0`.

## The oracle is a design decision — deliberately not pinned

The tests assert only that the error sentence is **absent**; they do not assert an exact replacement string. What is measured is that the current output is wrong. What is *chosen* is what the right output should be, and at least three answers are defensible:

- `"0 seconds"` / `"(0 secs)"` — matches hashcat's clamp, so the UI reads the same as `hashcat --status`.
- `"finishing"` — arguably truer; the mask really is done and only the next autotune is pending.
- `None` / empty — let the dashboard decide how to render "no estimate" rather than baking a display string into the agent.

A fix picking any of those passes. Only a fix that still lets the sentence reach `Time_Estimated` fails. There is a comment in the file saying so, and tightening the asserts once the project settles on a string is the natural follow-up.

## Adjacent, not asserted here

`guess_base_offset` / `guess_base_count` / `guess_base_percent` are hashcat's "mask N of M" (shown as `Guess.Queue`). `hashcatParser` reads only `estimated_stop`, `recovered_hashes` and `devices`, so honest file-level progress needs forwarding rather than new computation. One test documents that the data is present, but nothing here asserts we must surface it — that belongs with the #402 fix proper.

## Verification

Full fast gate, as CI runs it: **1516 passed, 2 skipped, 43 xfailed, 5 xpassed**. ruff, bandit and OpenAPI validation clean (the repo's pre-push hook ran all of it). The 5 xpasses are pre-existing on `v0.8.3-dev` and unrelated to this change.

Branched from and rebased onto current `v0.8.3-dev`; one commit, one file.

## Not verified

The chunking half of #402 is untouched here. I have not measured a multi-GPU agent, and have not confirmed the display behaviour end-to-end on a running server for a *long* single mask — the trace above is per-mask status output, not a dashboard screenshot.

### DIFF
--- a/tests/agent_unit/test_issue_xfail_eta_clamp.py
+++ b/tests/agent_unit/test_issue_xfail_eta_clamp.py
@@ -1,0 +1,201 @@
+"""xfail tests pinning the agent ETA bug from issue #402.
+
+#402 is chiefly about mask files bypassing chunking. This file covers the
+user-visible ETA symptom of the same root cause, which is independent of the
+planner and fixable on its own.
+
+hashcat's ``estimated_stop`` is the absolute epoch at which the CURRENT attack
+ends. With a mask file every line is its own attack, so the value resets per
+line and, on the terminal status line of each mask, equals the moment the mask
+finished. ``time_difference`` (install/hashview-agent/hashview-agent.py:533)
+subtracts ``datetime.now()`` and, for any delta <= 0, returns the sentence
+"The specified time is in the past." ``hashcatParser`` wraps that in parens as
+``Time_Estimated`` (:598) and the dashboard's ``_eta_text``
+(hashview/main/routes.py:170) prints whatever is between the parentheses, so an
+operator sees:
+
+    eta (The specified time is in the past.)
+
+hashcat itself never does this. At 100% progress its own status output reads
+``Time.Estimated...: <date> (0 secs)`` — it clamps the relative figure at zero.
+The tests below assert only that the error sentence does not reach the ETA
+field; they deliberately do not pin an exact replacement string. See the note
+above test_completed_attack_clamps_instead_of_reporting_an_error for why that
+choice is a design decision rather than something the measurement settles.
+
+Measured against hashcat v7.1.2-484-g64e1bff93 (single GPU, -m 0, three long
+masks, --status-timer=2). The countdown is accurate while a mask runs
+(+9s, +6s, +3s, +2s, 0s) and only the terminal line of each mask goes negative,
+so this is a sawtooth that repeats once per line of the file rather than an
+artefact of cheap masks. STATUS_JSON_TERMINAL below is a real captured line
+from that run.
+
+Each xfail asserts the DESIRED behaviour and is ``strict=True``: when the bug
+is fixed the test passes and strict-xfail turns that unexpected pass into a
+failure, as a reminder to drop the marker. Matches the convention in
+tests/agent_unit/test_issue_xfail_agent_robustness.py.
+"""
+import importlib.util
+import json
+import os
+import sys
+import time
+import types
+from pathlib import Path
+from unittest import mock
+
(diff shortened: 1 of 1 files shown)


bandrel:
Cross-referencing #428, which pins the server-side half of this sentinel (`_eta_text` in `hashview/main/routes.py` passes the sentence through to the dashboard ETA column, and `_eta_seconds` scores it 0). This PR clamps at the agent; that one keeps the server defensive for not-yet-upgraded agents and for `hc_status` rows already persisted. Different files, so no conflict.

Issue #412, opened 2026-08-21
Password history is dropped silently and unconditionally; no way to keep it for reuse analysis


bandrel:
## Summary

Password-history entries are dropped unconditionally on import, with no way to keep them. That is the right default, but it should be a choice, because history is exactly the data you need for password-reuse and rotation analysis.

Current behavior on `v0.8.3-dev` (`hashview/utils/utils.py:662`): any pwdump username containing `_history` is skipped, so both `COMPUTER$_history1` (correct — never crackable, pure noise) and `alice_history0` (arguably valuable) are discarded. #409, #410 and #411 propose extending that same drop to the `user:hash` and NetNTLM paths for consistency, which makes the question sharper rather than moot: after those fixes, history is dropped everywhere.

## Why keeping it can be the right call

An NTDS dump collected with `-history` is the only way to answer questions an assessment actually gets asked:

- Does this org rotate passwords, or increment them? `Summer2024!` → `Summer2025!` shows up only in history.
- Is the current password a reuse of a previous one for the same account?
- How effective was a forced org-wide reset — did people revert to a prior password?

Today that analysis is impossible in Hashview even when the operator deliberately collected the data, and the drop is silent — no count, no warning, nothing in the import summary saying "382 history entries discarded".

## Proposed shape

Split the two cases, since they are not the same decision:

- **Machine accounts** (trailing `$`, including their history) — keep dropping unconditionally. 120-char random secrets, no analytical value, pure denominator inflation.
- **User password history** — a per-import checkbox (default off, matching today's behavior), with the entries flagged rather than merged into the normal population so they can be excluded from crack-rate math but still queried for reuse.

At minimum, even without the toggle: report the count of dropped rows in the import result, so the operator knows the data was discarded rather than silently missing.

## Open questions

- Does a flag belong on the hashfile, on the individual `hashfile_hashes` row, or is a derived-at-query-time suffix match good enough? A row-level column is more honest but means a migration.
- Analytics needs an explicit rule either way: history entries must not count toward total accounts or crack rate, or every percentage on the page changes meaning depending on how the dump was collected.

Raising this as a design question, not a defect — filed so the deliberate-drop behavior is documented somewhere other than an unexplained `continue`.

i128:
[...]

Issue #411, opened 2026-08-21
NetNTLM import filters trailing $ but not $_history, so MACHINE$_history0 is imported


bandrel:
## Summary

The NetNTLM import filter only tests for a trailing `$`, so a machine-account name carrying a history suffix is not recognised.

`hashview/utils/utils.py:677`:

```python
elif file_type == 'NetNTLM':
    # First extracting usernames so we can filter out machine accounts
    if re.search(r"\$$", line.split(':')[0]):
        continue
```

`MACHINE$_history0::CORP:...` does not end in `$` and there is no `_history` clause on this branch (unlike the pwdump branch at `utils.py:662`), so it is imported and then uppercased to `MACHINE$_HISTORY0`.

Verified against `import_hashfilehashes` with `file_type='NetNTLM'`, `hash_type='5500'`, on `v0.8.3-dev` (a8c08dd).

## Impact

Low in practice, and I want to be upfront about that: NetNTLMv1/v2 hashes are challenge/response captures from the wire, and password history does not exist in that channel. I do not have a real-world capture that produces this, and it may not be reachable at all through normal Responder/ntlmrelayx output.

Filing it anyway because it is the same defect class as #409 and #410, the fix is the same one-line predicate swap, and leaving one of three branches inconsistent is how this family of bug regenerated in the first place. If the consensus is that it is unreachable, closing as wontfix is a fine outcome — but then the pwdump/NetNTLM asymmetry deserves a comment explaining why the two branches differ.

## Suggested fix

Route all three format branches through the single shared predicate proposed in #410, rather than three per-format regexes that drift apart.

i128:
fixed in https://github.com/hashview/hashview/pull/415

Issue #410, opened 2026-08-21
pwdump password-history filter is case-sensitive: COMPUTER$_HISTORY0 is imported as a real account


bandrel:
## Summary

The pwdump password-history filter is a case-sensitive substring test, so an uppercased dump slips straight through.

`hashview/utils/utils.py:662`:

```python
if re.search(r"\$$", line.split(':')[0]) or "_history" in line.split(':')[0]:
    continue
```

`"_history" in ...` matches lowercase only. `WIN10$_HISTORY0` does not end in `$` and does not contain the lowercase literal, so it is imported as a normal account.

## Reproduction

pwdump / NTLM (1000):

```
alice:1001:aad3b435b51404eeaad3b435b51404ee:8846f7eaee8fb117ad06bdd830b7586c:::
WIN10$:1002:aad3b435b51404eeaad3b435b51404ee:8846f7eaee8fb117ad06bdd830b7586d:::
WIN10$_history0:1003:aad3b435b51404eeaad3b435b51404ee:8846f7eaee8fb117ad06bdd830b7586e:::
CORP\WIN10$_history1:1004:aad3b435b51404eeaad3b435b51404ee:8846f7eaee8fb117ad06bdd830b7586f:::
WIN10$_HISTORY0:1005:aad3b435b51404eeaad3b435b51404ee:8846f7eaee8fb117ad06bdd830b75870:::
```

Imported: `alice`, **`WIN10$_HISTORY0`**. Everything else is correctly dropped, including the domain-prefixed variant.

Verified against `import_hashfilehashes` on a scratch SQLite app, `v0.8.3-dev` (a8c08dd).

## Impact

Lower than #409 — impacket's `secretsdump.py` emits `_history` lowercase, so the common tooling path is unaffected. It bites when the dump has been uppercased in preprocessing, or comes from a tool that cases the suffix differently. One machine-account history entry per computer object per rotation is enough to skew a hashfile's crack rate, and nothing filters at report time.

## Suggested fix

Case-insensitive, and anchored rather than a bare substring:

```python
_HISTORY_RE = re.compile(r'_history\d*$', re.IGNORECASE)
```

Anchoring is a small behavior change worth calling out: today a real user named `bob_historyclub` is silently dropped, because the check looks for `_history` anywhere in the name. The AD format is always `<name>_history<n>` at the end, so anchoring fixes that false positive at the same time.

## Related

[...]

PR #415, opened 2026-08-21
Filter machine accounts and NTLM password history in every import format (#409, #410, #411)

### LINKED ISSUE #409: user:hash import path never filters machine accounts or password history (COMPUTER$ and COMPUTER$_history1 both imported)
## Summary

Machine-account and password-history filtering on import is implemented per-format, and only two of the formats have it. The `user:hash` format (`file_type == 'user_hash'`, API `file_format` 4) has **no filter at all**, so `COMPUTER$` and `COMPUTER$_history1` are both imported as if they were real user accounts.

There are exactly two machine-account filters in the entire codebase, both in `import_hashfilehashes` (`hashview/utils/utils.py`):

- `utils.py:662` — pwdump: `if re.search(r"\$$", line.split(':')[0]) or "_history" in line.split(':')[0]`
- `utils.py:677` — NetNTLM: `if re.search(r"\$$", line.split(':')[0])`

The `user_hash` branch (`utils.py:634-653`) extracts `username = line.split(':')[0]` and imports unconditionally.

## Reproduction

Importing this as `user:hash` / NTLM (1000):

```
alice:8846f7eaee8fb117ad06bdd830b7586c
WIN10$:8846f7eaee8fb117ad06bdd830b7586d
WIN10$_history0:8846f7eaee8fb117ad06bdd830b7586e
```

yields all three rows in `hashfile_hashes`. The same content in pwdump format correctly yields only `alice`.

Verified by calling `import_hashfilehashes` directly against a scratch SQLite app on `v0.8.3-dev` (a8c08dd).

## Why it matters

This is the realistic path for the bug, not a contrived one. `user:hash` is what you get whenever an NTDS dump is cut down to `username:nthash` — a very common preprocessing step — and it is a first-class option in both the UI file-type dropdown (`hashview/jobs/forms.py:49`) and the API (`hashview/api/routes.py:1322`, format 4).

Because nothing filters at report time — the two lines above are the only filtering anywhere in the app — every machine account that lands via `user:hash` permanently skews that hashfile's analytics: crack rate, total-accounts count, and password-reuse counts. Machine account passwords are 120-character random strings that will never crack, so they inflate the denominator and depress the reported recovery percentage. Password-history entries inflate it further and double-count reuse.

## Suggested fix

Extract the predicate into one helper (something like `is_machine_or_history_account(username)`) and apply it in every branch that produces a username, rather than repeating the regex per format. See #410 and #411 for the two other formats where the existing checks are incomplete.

## Scope notes

- `hash_only` needs nothing — it stores no username.
- `shadow` needs nothing — Unix shadow files have no AD machine accounts or NTLM history.
- `kerberos` (`utils.py:668-673`) can legitimately carry machine-account principals, but password history does not exist in ticket captures. Worth a look, but it is a different question from this one and I have not investigated it.

## What I have not verified

I have not checked whether any deployment has already imported machine accounts this way, so I do not know whether a cleanup/migration path is also needed for existing hashfiles.

bandrel:
Fixes #409, #410, #411.

## What was wrong

Machine-account and password-history filtering on import was implemented per-format, and only two of the five format branches in `import_hashfilehashes` had it. Verified by running the real parser against synthetic dumps:

| Format | Input | Imported before |
|---|---|---|
| pwdump | `alice`, `WIN10$`, `WIN10$_history0`, `CORP\WIN10$_history1`, `WIN10$_HISTORY0`, `krbtgt` | `alice`, `krbtgt`, **`WIN10$_HISTORY0`** |
| user:hash | `alice`, `WIN10$`, `WIN10$_history0` | **all three** |
| NetNTLM | `MACHINE$_history0` | **imported** |

- **#409** — `user:hash` had no filter at all. This is the realistic path: it is what you get whenever an NTDS dump is cut down to `username:nthash`, and it is a first-class option in both the UI dropdown and the API (format 4).
- **#410** — the pwdump check was `"_history" in username`, a case-sensitive substring test, so `WIN10$_HISTORY0` slipped through.
- **#411** — the NetNTLM check was trailing-`$` only, so `MACHINE$_history0` slipped through. Low real-world reachability (network captures carry no password history), fixed for consistency.

Nothing filters at report time — those two lines were the only filtering anywhere in the app — so each leaked row permanently skews that hashfile's account count and crack rate. Machine-account passwords are 120-char random secrets that will never crack, so they inflate the denominator and depress the reported recovery percentage.

## What changed

The three per-format regexes are replaced by one predicate in `hashview/utils/utils.py`:

```python
_HISTORY_SUFFIX_RE = re.compile(r'_history\d*$', re.IGNORECASE)

def is_machine_or_history_account(username):
    name = (username or '').strip()
    return name.endswith('$') or bool(_HISTORY_SUFFIX_RE.search(name))
```

applied in the pwdump, `user_hash`, and NetNTLM branches.

Two judgement calls worth a reviewer's attention:

1. **The history suffix is now anchored** (`_history<n>` at end of name) instead of a substring match anywhere. This is a deliberate behavior change beyond the reported bugs: the old check silently dropped a real account named `bob_historyclub`. The AD format always puts the suffix last, so anchoring costs nothing and fixes that false positive. Existing tests still pass — the property-test `USERNAME` strategy generates no underscores, so nothing depended on the looser match.
2. **Filtering is scoped to the NTLM family** (`NTLM_FAMILY_HASH_TYPES`) rather than applied to every `user:hash` import. `user:hash` is a generic format used for MD5 web-app dumps and the like, and machine-account/history semantics are AD-specific — dropping a trailing-`$` username out of an MD5 dump would be silent data loss, not a fix. pwdump and NetNTLM are unaffected by the scoping since both are inherently NTLM (`validate_pwdump_hashfile` already rejects any non-1000 hash type). Happy to widen this if reviewers disagree.

The filter also runs *before* `import_hash_only`, so a filtered row leaves no orphaned ciphertext in `hashes` that would still get cracked. There is a test for that specifically.

## Tests

New `tests/unit/test_machine_account_filter.py`: 22 tests — a parametrized table for the predicate (including the anchoring false positive and the whitespace case), one regression test per issue, the non-NTLM scoping case, and the no-orphan-hash-rows case. Written before the fix and confirmed red first (`ImportError` on the missing helper).

Each of the three guards was then mutation-tested by reverting it individually and confirming the matching tests fail:

- remove the `user_hash` guard → `test_user_hash_ntlm_filters_machine_accounts_and_history`, `test_filtered_rows_do_not_create_hash_rows` fail
- make the regex case-sensitive and unanchored → 5 tests fail (`test_pwdump_filters_uppercase_history`, `test_pwdump_keeps_account_with_history_inside_the_name`, 3 predicate cases)
- revert NetNTLM to trailing-`$` only → `test_netntlm_filters_machine_account_history` fails

## Verification

- `pytest tests/unit tests/security tests/agent_unit` — **1536 passed, 2 skipped, 39 xfailed, 5 xpassed**, 0 failures
- `pylint --rcfile=.pylintrc hashview/ hashview.py` — **9.63/10**, unchanged from the pre-change run
- `ruff check hashview/ hashview.py` — clean
- CHANGELOG entry added under Fixed

No migration, no schema change, no API change.

## Not addressed here

Existing hashfiles that already imported machine accounts are untouched — this only fixes the import path, so a cleanup for already-polluted data would be separate work. #412 tracks the related design question of whether *user* password history should be droppable-by-choice rather than unconditional, since it is the data you need for password-reuse analysis; this PR keeps the current drop-everything behavior and only makes it consistent. The `kerberos` branch can carry machine-account principals and is untouched — I have not investigated it.


bandrel:
Follow-up commit 5019410 after two independent review passes (spec compliance, then code quality). Both confirmed #409/#410/#411 are genuinely fixed and reproduced the verification numbers, but each surfaced a real gap:

**DCC2 was still leaking.** `hash_only` + hash type 2100 *does* store a username — `$DCC2$<iter>#<user>#<hash>`, extracted at the old `utils.py:654` — so it can carry an AD machine account, and 2100 is UI-selectable. My scope note on #409 ("hash_only needs nothing, it stores no username") was wrong for that one mode. Now filtered, with a test.

**The `str(hash_type)` coercion was untested.** It is load-bearing rather than defensive: the API upload route is declared `<int:hash_type>` and passes the value straight through, so without it the filter silently no-ops on the API path. Every test passed a string, i.e. only the form path. Added the int case.

Also in this commit: narrowed the hash-type set to the modes actually reachable through the gate (`{1000, 2100, 3000}` — the NetNTLM modes arrive as `file_type` NetNTLM, which filters unconditionally, so they were unreachable set members), made it private to match the file's `_`-prefix convention for module internals, dropped two comments that went stale once nothing was "extracted first", and corrected the CHANGELOG's "every import format" to what is actually true (kerberos ticket imports are still unfiltered, and that is now stated).

Both new guards were mutation-tested individually: removing the DCC2 filter fails `test_hash_only_dcc2_filters_machine_accounts`, and dropping the `str()` fails `test_user_hash_filters_when_hash_type_is_an_int`.

Re-verified locally: 1538 passed / 2 skipped / 39 xfailed, 0 failures; pylint 9.63 unchanged; ruff clean.

Corrections to the PR description above: `import_hashfilehashes` has six format branches, not five, and NetNTLM had a partial filter rather than none.

i128:
Good fix, thanks

Issue #171, opened 2026-06-02
Tracking: audit ownership/admin enforcement across remaining blueprints


bandrel:
The recent IDOR fix only covered `hashview/jobs/routes.py` — every endpoint that takes a resource-id path arg there now goes through `_authorize_job_or_abort()`. The same class of bug very likely exists elsewhere: any route that accepts a `<resource>_id` path parameter and only requires `@login_required` allows cross-tenant access.

Blueprints not yet audited:

- [ ] `hashview/customers/routes.py`
- [ ] `hashview/hashfiles/routes.py`
- [ ] `hashview/tasks/routes.py`
- [ ] `hashview/task_groups/routes.py`
- [ ] `hashview/rules/routes.py`
- [ ] `hashview/wordlists/routes.py`
- [ ] `hashview/notifications/routes.py`
- [ ] `hashview/users/routes.py` (profile edits, password resets, API key regen)
- [ ] `hashview/agents/routes.py` (verify scoping on agent-controlled cookies)
- [ ] `hashview/api/routes.py` — `is_authorized()` only validates the presence/correctness of an API key; check whether `/v1/jobs/<id>`, `/v1/tasks/<id>`, `/v1/hashfiles/<id>`, etc. actually scope returned rows to the calling user

`_authorize_job_or_abort` in `hashview/jobs/routes.py` is a template for the pattern, but a cross-blueprint variant (probably in `hashview/utils/`) is the cleaner long-term shape.

Adding an e2e test analogous to `test_job_idor_access_denied_for_other_user` for each resource type would catch regressions.

bandrel:
Worked through this checklist against the current `v0.8.3-dev`. Two findings, and the second one is the reason I am commenting rather than closing.

**The premise needs correcting.** This issue describes the jobs-blueprint fix as introducing a shared `_authorize_job_or_abort()` helper. No such helper exists anywhere in the tree (`grep -rn` finds nothing). What `jobs/routes.py` actually does is an inline `current_user.admin or X.owner_id == current_user.id` check per mutating route — e.g. `jobs_delete` at `hashview/jobs/routes.py:821`, `jobs_stop` at `:969`. So "extract the helper and apply it everywhere" is not the shape of the remaining work.

[...]

Issue #401, opened 2026-08-19
API: no endpoints for task groups — groups can only be built in the web UI


bandrel:
## Summary

Task groups exist only in the web UI. The `task_groups` blueprint (`hashview/task_groups/routes.py`) covers the whole lifecycle — create (`/task_groups/add`, line 53), assign (`/task_groups/assigned_tasks/<id>/add_task/<id>`, line 149), reorder (`promote_task`/`demote_task`, lines 180 and 209), delete (line 238) — and every one of those routes is `@login_required` HTML. There is no `/v1` counterpart: grepping `hashview/api/routes.py` for `task_group` finds only the internal `_cancel_task_group` helper (line 252).

That leaves a hole in an otherwise scriptable path. The API can create the pieces — `POST /v1/tasks/add` (line 1188), `POST /v1/rules/add/<name>` (line 707), `POST /v1/wordlists/add/<name>` (line 863), `POST /v1/jobs/add` (line 1012) — but not the grouping that organises them.

## Impact

Two things break for an API-only client.

**Curated task sets can't be assembled.** I generate tasks in bulk from a config file (hashview-utils), which works fine, but grouping them is then hand-work in the browser. My current case is 7 mask-file tasks that belong in one "Corporate Masks 8-14" group; the create half is scripted and the group half is clicking.

**`POST /v1/jobs/add` can't be pointed at a chosen task set.** It builds job tasks from the "top 10 effective tasks" for the hashfile's hash type (`routes.py:1054-1063`) and offers no parameter to override that. Combined with the missing group endpoints, there is no way to say "run exactly these tasks against this hashfile" over the API — the one place that picks tasks for you picks them by past effectiveness, and the one place a human curates a set is UI-only.

## Suggested fix

Four endpoints in `hashview/api/routes.py`, mirroring the auth and envelope of `v1_api_post_add_job` (`is_authorized(user=True, agent=False, ...)`, resolve the caller by `uuid` api_key cookie, return the `{status, type, msg}` JSON shape):

- `GET /v1/task_groups` — list groups with `tasks` already parsed into an id array, rather than making clients re-parse the stored string.
- `POST /v1/task_groups/add` — `{"name": ..., "tasks": [id, ...]}`, creating the group and its ordered membership in one call. Validate every id against `Tasks` and `404`/`400` on an unknown one.
- `POST /v1/task_groups/<int:task_group_id>/tasks` — set or append membership, preserving order, so a group can be built incrementally the way the UI does.
[...]

Issue #399, opened 2026-08-18
API: no way to replace the contents of an existing rule or wordlist (no PUT /v1/rules/<id> or /v1/wordlists/<id>)


bandrel:
## Summary

The v1 API can **create** rules and wordlists but has no way to **replace the contents of an existing one**. There is no `PUT /v1/rules/<int:rule_id>` and no `PUT /v1/wordlists/<int:wordlist_id>` — the only write routes are `POST /v1/rules/add/<rule_name>` (`hashview/api/routes.py:707`) and `POST /v1/wordlists/add/<wordlist_name>` (`routes.py:863`), both of which append a brand-new row.

The web UI is only half a workaround: rules have an in-place content editor (`GET/POST /rules/edit/<int:rule_id>`, `hashview/rules/routes.py:130`, which rewrites `rule.path` and recomputes `size`/`checksum`), but **wordlists have no edit route at all** — `hashview/wordlists/routes.py` offers add, delete, download and update-dynamic, and nothing that swaps a static wordlist's file.

## Impact

Delete-then-recreate is not an equivalent workaround, for three separate reasons:

1. **Delete is blocked precisely when you need it.** Both delete paths refuse when the resource is referenced by a task — `Tasks.query.filter_by(rule_id=rule.id)` (`rules/routes.py:195`) and the `Tasks.wl_id` check (`wordlists/routes.py:196-201`). The rules and wordlists worth refreshing are the ones already wired into tasks, so they can never be removed and re-added.
2. **The id changes.** A recreate mints a new `Rules`/`Wordlists` row, so every `Tasks.rule_id` / `Tasks.wl_id` still points at the old row, and any client holding an id has to re-resolve. Name-based re-resolution is itself unreliable today because the add endpoints don't dedupe names (see #397/#398).
3. **The API can't delete anything yet.** `DELETE /v1/rules/<id>` and `DELETE /v1/wordlists/<id>` don't exist — that's #397 and #398. Until those land, an API-only client cannot even attempt the workaround.

The concrete case is routine catalog maintenance: refreshing a curated rule file or swapping in a newer wordlist build across instances, from `migrate_catalog.py` in hashview-utils. Today that requires web-UI access for rules and is simply impossible for wordlists.

[...]

Issue #395, opened 2026-08-18
setup: default wordlist/rule seeding fails with EXDEV (os.replace) when control/ is a mount — blocks the volume fix for #383


bandrel:
## Summary

`add_default_static_wordlist` and `add_default_rules` install their seed files with `os.replace()`:

```python
# hashview/setup/__init__.py:81-83
os.system('gzip -d -k install/rockyou.txt.gz')
wordlist_path = 'hashview/control/wordlists/rockyou.txt'
os.replace('install/rockyou.txt', wordlist_path)

# hashview/setup/__init__.py:62-64
os.system('gzip -d -k install/best64.rule.gz')
rules_path = 'hashview/control/rules/best64.rule'
os.replace('install/best64.rule', rules_path)
```

`os.replace()` is `rename(2)`, which cannot cross filesystems. As soon as `control/` is a mount — a Docker volume, a bind mount, a separate partition — `install/` and `control/` are different devices and both calls fail with `EEXDEV`.

## Reproduction

Add the volumes that `control/` needs in order to survive a container rebuild, then boot fresh:

```yaml
services:
  app:
    build: .
    volumes:
      - wordlists:/hashview/control/wordlists
      - rules:/hashview/control/rules
```

First boot:

```
[INFO ] Adding Default Static Wordlist.
[ERROR] Adding Default Static Wordlist failed.
Traceback (most recent call last):
  File "/hashview/__init__.py", line 103, in setup_defaults_if_needed
    add_default_static_wordlist(db)
  File "/hashview/setup/__init__.py", line 83, in add_default_static_wordlist
    os.replace('install/rockyou.txt', wordlist_path)
OSError: [Errno 18] Cross-device link: 'install/rockyou.txt' -> 'hashview/control/wordlists/rockyou.txt'

[INFO ] Adding Default Rules.
[ERROR] Adding Default Rules failed.
OSError: [Errno 18] Cross-device link: 'install/best64.rule' -> 'hashview/control/rules/best64.rule'
```

## Why it's easy to miss

Both call sites in `hashview/__init__.py:100-105` and `:126-131` wrap the seeding in `try/except Exception` + `logger.exception`, so the app boots healthy and serves normally. You only notice later that there is no Rockyou wordlist and no Best64 rule — the failure is a log line, not a startup error.

Because `default_static_wordlist_need_added()` / `default_rules_need_added()` stay `True`, it also retries and fails identically on every single boot.

## Why it matters beyond an exotic setup

[...]

PR #417, opened 2026-08-21
docs: remediation roadmap for the open-issue backlog

### LINKED ISSUE #355: Cracked-hash import rejects hashcat $HEX[...] plaintexts

### LINKED ISSUE #374: API: report how many founds are NEW on import (counterpart to instacracked)
## Summary

`instacracked` tells you how many uploaded hashes this instance *already had plaintext for*. The missing counterpart is the opposite outcome: how many **founds are new** — that is, how many hashes this import moved from uncracked to cracked, contributing plaintext the instance did not previously hold.

The distinction that matters is about **recoveries, not hash rows**. A hash being new to the `hashes` table is uninteresting on its own; what an operator wants to know after bringing results in is "how much did I actually teach this server?"

## Terminology

For an import that carries plaintext, every submitted line falls into exactly one bucket:

| Bucket | Meaning |
|---|---|
| `new_founds` | The hash existed uncracked; this import supplied plaintext and flipped it to cracked. **The number being requested.** |
| `known_founds` | The hash was already cracked here; the plaintext adds nothing. (This is the `instacracked` case.) |
| `conflicting_founds` | The hash was already cracked here but with a *different* plaintext. Should not happen; if it does it means a collision, a mode mismatch, or bad input, and it should be surfaced rather than silently overwritten. |
| `unmatched` | Plaintext supplied for a ciphertext this instance doesn't have. Nothing to attach it to. |

`new_founds + known_founds + conflicting_founds + unmatched` should equal the number of input lines, so the response is self-checking.

## Where this belongs

This is a property of an import that carries plaintext, so it belongs on the existing founds-import endpoint `POST /v1/hashes/import/<hash_type>` (hashview/api/routes.py:1872) — not on `POST /v1/hashfiles/upload/...`, which only ever receives ciphertext and therefore can never produce a new found.

(An earlier revision of this issue proposed `new_hashes` / `existing_hashes` on the hashfile-upload route, counting rows new to the `hashes` table. That measures the wrong thing: hashes are inserted with `cracked=0`, so "new to the DB" says nothing about recoveries. Superseded by the bucket definitions above.)

## Why it's worth reporting

- It is the only way to tell a redundant re-import from one that added real knowledge. Both currently return a success body that looks identical.
- `new_founds` is precisely the set that will drive future `instacracked` results and future `(DYNAMIC) All Recovered Passwords` growth, so it quantifies what the import bought you going forward.
- `conflicting_founds` is a genuine correctness signal that no current endpoint surfaces.

## Implementation note

The per-hash outcome is already determined during import; it just isn't propagated. `import_hash_only()` (`hashview/utils/utils.py:544`) returns a bare id from both its found and created branches, and the crack-marking path in `v1_api_post_jobtask_crackfile_upload` similarly discards whether the row it updated had been cracked before. Counting these requires no additional queries, only returning the outcome alongside the id.

## Related

- #372 — closed; the founds-import endpoint already exists as `POST /v1/hashes/import/<hash_type>`.

## Current response

The route today returns a bare acknowledgement with no counts at all:

```json
{"msg": "OK", "status": 200, "type": "message"}
```

So an operator importing 34,447 founds cannot tell how many were newly recovered, how many were already known, or how many matched nothing in this instance. Note also that lines whose ciphertext isn't present are silently skipped (the `if record:` branch simply doesn't fire), which is exactly the `unmatched` bucket above — currently invisible.

bandrel:
Docs-only. Adds `docs/plans/2026-08-21-open-issue-remediation.md`.

## What this is

A triage of every issue that was open on the board (77 at the time of writing), verified against the code on `v0.8.3-dev` @ a8c08dd rather than against the issue text, sequenced into waves with the couplings that decide ordering written down.

It is deliberately a portfolio roadmap and not seventy implementation plans — each wave gets its detailed per-task plan when it starts, because plans written months ahead of the work go stale before anyone opens an editor.

## The parts worth reading even if you skip the rest

**Nine open PRs already fix issues on the board.** Landing them is the highest-yield item here and costs review time, not engineering time: #340, #354, #356, #358, #391, #400, #408, #415, #416.

**Two corrections that affect what gets closed:**
- PR #356's description claims it adds founds-import count fields and an `openapi.yaml` update. Its actual diff touches only `hashview/utils/utils.py` and two test files. It fixes #355 correctly but does **not** fix #374 — don't close #374 on merge.
- PR #367 is superseded; the `control/tmp` cleanup already landed at `api/routes.py:174-189` plus three call sites. #226 is closed citing that code.

**Orderings that are load-bearing, not preference:**
- #395 (EXDEV on `os.replace` when `control/` is a mount) blocks both #399's safe file swap and #383's volume fix.
- #384 (mask store) makes #402's chunking bug *reachable* — `chunking.py:117-126` gives a path-shaped mask a keyspace of 1, so it never chunks. Today no mask file can reach an agent, so it's theoretical. The interim guard has to ship with #384, not after it.
- #374's bucket semantics before #376, since the hashcat delegation produces exactly those buckets.
- The agent must learn to read real HTTP status codes before #221 flips the server's 200s to 401/403, or every deployed agent breaks.

**Five items are design questions, not defects,** and are called out so an implementer doesn't pick one up: #350 (flask-smorest go/no-go), read-side authorization, #412 (history retention), #401 (task-group storage), #123/#314 (scope down or close).

**One finding came out of the #171 audit that isn't tracked anywhere yet.** Mutation routes are uniformly owner-gated across every blueprint — that checklist is genuinely clean. Reads are uniformly *not*, and `hashfiles_download` (`hashfiles/routes.py:240-241`) has only `@login_required` while its `cracked`/`plains` formats return recovered plaintext for a guessable integer id. `tests/e2e/test_security.py:184` says the shared-view model is deliberate, so whether that's acceptable depends on whether all authenticated users are trusted across all client engagements. Left as a question on #171 rather than changed unilaterally.

## Board hygiene done alongside

Closed as already fixed, each with a code citation: #92, #129, #130, #141, #164, #226. Closed as an environment thread: #116. Commented rather than closed: #171 and PR #367.

## Verification

Docs-only, no code touched. Statuses in the document were verified by reading the tree; the handful that were not are explicitly labelled "not independently verified in this pass" so nobody treats them as checked.

### DIFF
--- a/docs/plans/2026-08-21-open-issue-remediation.md
+++ b/docs/plans/2026-08-21-open-issue-remediation.md
@@ -12,7 +12,7 @@ Every status below was verified by reading the code on `v0.8.3-dev`, not by trus
 
 ## Sequencing principles
 
-1. **Merge before build.** Nine open PRs already implement fixes for open issues. Landing them removes work from the board at review cost only, and several later waves touch the same files.
+1. **Merge before build.** Ten open PRs already implement fixes for open issues. Landing them removes work from the board at review cost only, and several later waves touch the same files.
 2. **Correctness before features.** A wrong number on the analytics page is worse than a missing one, because nobody knows to distrust it.
 3. **One PR per code region, not one PR per issue.** Where several issues live in the same function, fixing them separately means three reviewers reading the same diff three times and two of them rebasing.
 4. **Decide the design questions first.** Four items on the board are questions, not defects. Implementing them before the question is answered is how you get a feature nobody wanted.
@@ -21,7 +21,7 @@ Every status below was verified by reading the code on `v0.8.3-dev`, not by trus
 
 ## Wave 0 — Land what is already written
 
-Nine PRs are open against issues on the board. No new code required; this is review-and-merge capacity, and it is the single highest-yield thing on this roadmap.
+Ten PRs are open against issues on the board. Nine need no new code; this is review-and-merge capacity, and it is the single highest-yield thing on this roadmap.
 
 | PR | Closes | Notes |
 |---|---|---|
@@ -36,7 +36,7 @@ Nine PRs are open against issues on the board. No new code required; this is rev
 | #416 | #413, #414 | Server-side list filters; paginate and sort `/rules`. |
 | #420 | #226 (reopened) | Makes the `control/tmp` cleanup actually run; see below. |
 
-Nine PRs plus #420. Three things to fix while merging:
+Three things to fix while merging:
 
 - **PR #356's description overstates its diff.** The body claims it adds `verified`/`updated`/`unmatched`/`count` response fields and an `openapi.yaml` update. The actual diff touches only `hashview/utils/utils.py` and two test files — no route, no spec. It fixes #355 correctly; it does **not** fix #374. Do not close #374 on merge.
 - **PR #400 references #377 but does not fix it.** It fixes the job-collision bug (one shared wordlist row, concurrent jobs overwriting each other's crawl), which is worth merging. But `words = crawl_website_keywords(target, settings)` is still unguarded — it appears in the diff only as context — and the `try`/`except OSError` blocks it adds wrap the tmp-file write and move, not the crawl. A raising crawl still serves the previous run's file. Leave #377 open.
@@ -44,7 +44,7 @@ Nine PRs plus #420. Three things to fix while merging:
 
   The general lesson is worth carrying into the other waves: a green test that drives a helper directly can assert a contract the request path never exercises. Prefer a real request.
 
-**Exit criteria:** all nine remaining merged or explicitly rejected; #218, #226, #353, #355, #357, #105, #409, #410, #411, #413, #414 closed. #377 stays open.
+**Exit criteria:** all ten merged or explicitly rejected; #218, #226, #353, #355, #357, #105, #409, #410, #411, #413, #414 closed. #377 stays open.
 
 ---
 
--- a/docs/plans/2026-08-21-open-issue-remediation.md
+++ b/docs/plans/2026-08-21-open-issue-remediation.md
@@ -34,14 +34,17 @@ Nine PRs are open against issues on the board. No new code required; this is rev
 | #408 | #105 | e2e coverage for every top-level page. |
 | #415 | #409, #410, #411 | Machine-account / history filtering. Green, awaiting review. |
 | #416 | #413, #414 | Server-side list filters; paginate and sort `/rules`. |
+| #420 | #226 (reopened) | Makes the `control/tmp` cleanup actually run; see below. |
 
-Three things to fix while merging:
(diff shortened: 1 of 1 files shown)


bandrel:
Updated after a second sweep — two corrections to the document, both found by verifying claims I had taken on trust the first time.

**PR #400 does not fix #377.** The Wave 0 table originally listed it as closing #377. It doesn't: `words = crawl_website_keywords(target, settings)` is still an unguarded call (it appears in that diff only as context), and the `except OSError` blocks the PR adds wrap the tmp-file write and move, not the crawl. A raising crawl still serves the previous run's wordlist. #400 fixes the job-collision bug and is worth merging on that basis; #377 stays open. Flagged on #400 directly so it isn't auto-closed.

**The 19 unverified issues are now verified, and the board holds up.** The first pass labelled a batch "not independently verified" — those have now been checked one at a time against the code. Only #361 was resolved (the phosphor redesign's fluid main pane plus collapsible sidebar satisfies both halves of it); it is closed. The other eighteen are genuinely open with the gap located — `config.py:19-24` still concatenates the DB password into the URI with no `quote_plus` (#57), `docker-compose.yml:25` still publishes 3306 (#394), `setup/__init__.py:64,83` still call bare `os.replace` (#395), `add_default_tasks` still hardcodes `wl_id` 2/3 (#396).

Also noted: PR #367 is now closed as superseded, but the `control/tmp` cleanup that replaced it shipped with **no regression test** — nothing asserts the tmp listing is unchanged across a download, so the `call_on_close` hook could be removed with CI staying green. #367's three tests are worth cherry-picking.

PR #419, opened 2026-08-21
fix(githooks): pre-push scanner missed secrets on new-branch pushes

### LINKED ISSUE #418: pre-push hook passes rev-list args to git diff: secrets on a new-branch push are never scanned
`.githooks/pre-push` computes its scan range with `git rev-list` syntax but then feeds those same arguments to `git diff`. On a **new-branch push** that produces a *combined* diff (`diff --cc`) instead of the intended "lines added by this push", which breaks the scanner in both directions. The false-negative direction is the serious one: a commit containing an AWS key and a plaintext password passes the scan with `sensitive-content scan clean.`

## Cause

For a branch the remote doesn't have yet, the hook builds rev-list-style args (line 63):

```bash
if [ "$remote_sha" = "$zero" ]; then
    # New branch: scan everything reachable from local_sha not already on any remote.
    range_args=("$local_sha" --not --remotes)
else
    range_args=("${remote_sha}..${local_sha}")
fi
```

Those args are correct for `git rev-list` (line 69) and `git log` (line 80). They are then reused verbatim for two `git diff` invocations:

```bash
diff=$(git diff --unified=0 --no-color "${range_args[@]}" 2>/dev/null)          # line 73
changed_files=$(git diff --name-only --diff-filter=AM "${range_args[@]}" 2>/dev/null)  # line 76
```

`git diff` has no notion of `--not`/`--remotes` as range exclusion. Given multiple revisions it emits a combined diff, one column per extra revision:

```
diff --cc clean.py
@@@@@@ -1,0 -1,0 -1,0 -1,0 -1,0 +1,2 @@@@@@
+++++def add(a, b):
```

Combined-diff semantics are "how this differs from *each* parent," and hunks uninteresting relative to some parent are omitted. Neither property matches "what this push adds." The `2>/dev/null` hides any complaint, so the failure is silent.

The number of remote-tracking refs decides the outcome, which is why this varies per clone — a clone with several remotes/forks behaves differently from a fresh one.

## Impact 1 (serious): real secrets are not scanned

Reproduction, using the actual hook, against a repo with a remote branch older than the pushed content:

```bash
mkdir -p /tmp/hookrepro && cd /tmp/hookrepro
git init -q --bare upstream.git && git clone -q upstream.git clone && cd clone
git config user.email t@t.test && git config user.name T

printf 'print("hello")\n' > app.py
git add app.py && git commit -q -m initial
git branch -M main && git push -q origin main

# a fork/dev branch pushed before the later content existed (myusif/*, birkep/* here)
git checkout -q -b oldfork main && git push -q origin oldfork

git checkout -q main
printf 'print("hello")\nrequests.post("https://api.pushover.net/1/messages.json")\n' > app.py
git add app.py && git commit -q -m "add pushover notify" && git push -q origin main

# new branch containing an obvious secret
git checkout -q -b feature/x main
printf 'AWS_KEY = "AKIAIOSFODNN7EXAMPLE"\npassword = "sup3rs3cret_value"\n' > secret.py
git add secret.py && git commit -q -m "oops"
git fetch -q origin

mkdir -p .githooks && cp /path/to/hashview/.githooks/pre-push .githooks/
SHA=$(git rev-parse HEAD)
printf 'refs/heads/feature/x %s refs/heads/feature/x %s\n' "$SHA" "$(printf '0%.0s' {1..40})" | ./.githooks/pre-push origin ./upstream.git
```

Output:

```
pre-push: sensitive-content scan clean.
```

Both the `AWS access key ID` and `Generic secret assign` patterns would have matched had the diff been computed correctly — `git diff main...HEAD` shows the lines plainly. The scan simply received an empty haystack. A secret-scanning hook that reports "clean" on `AKIAIOSFODNN7EXAMPLE` is worse than no hook, because it is trusted.

`changed_files` (line 76) is built from the same broken arguments. In practice it still works, because a combined diff names the changed files even when it omits their hunks -- I verified `.env` additions are still flagged on a new-branch push. It should be corrected anyway for the same reason (`git diff` is not being given a range it understands), but it is not currently a hole.

## Impact 2 (noise): pre-existing lines reported as added

The other direction: with several remote refs, combined-diff columns surface lines that are already upstream and untouched by the branch. Observed on PR #416, whose diff adds no URLs at all:

```
[URL]
  619:+    ++ +                    +                            ++  ++            response = requests.post('https://api.pushover.net/1/messages.json', ...)
```

That `api.pushover.net` line is already on `origin/main` and `origin/v0.8.3-dev`. The `+ ++ +` gutter is the combined-diff column layout leaking into the report; the line number is meaningless. This trains people to bypass with `--no-verify`, which skips the entire gate — ruff, bandit, OpenAPI, and the test suite included.

## Suggested fix

`git log -p` accepts rev-list arguments natively and emits per-commit patches, which is exactly the intended semantics. Replacing both `git diff` calls keeps the existing range logic untouched:

```bash
diff=$(git log -p --unified=0 --no-color --format='' "${range_args[@]}" 2>/dev/null)
changed_files=$(git log --name-only --diff-filter=AM --format='' "${range_args[@]}" 2>/dev/null | sort -u)
```

`git log -p` skips merge commits by default, which is correct here: a merge introduces no content that isn't already in the individual commits within the range.

Worth doing alongside:

- Drop `2>/dev/null` on these, or check the exit status, so a future argument mismatch fails loudly instead of degrading to "clean".
- Fail closed: if `commits` is non-empty but the computed diff is empty, that is a contradiction (commits always change something) and should be treated as an error rather than a pass.
- Add regression coverage. There is currently no test for `.githooks/pre-push`; a shell test that plants a known-bad string in a new-branch push and asserts the hook exits non-zero would have caught this.

bandrel:
Fixes #418.

## Why

`.githooks/pre-push` computed its scan range with `git rev-list` syntax and then passed those same arguments to `git diff`. For a new branch that range is `<sha> --not --remotes`, which `git diff` does not understand as range exclusion — it treats the extra revisions as further sides to compare and emits a *combined* diff. Combined-diff semantics are "how this differs from each parent", which is not "what this push adds": it drops real additions and attributes untouched upstream lines to the push. The `2>/dev/null` on the call kept the mismatch quiet.

The result was a false negative in a secret scanner. On a new-branch push carrying an AWS access key ID and a plaintext password, the hook printed:

```
pre-push: sensitive-content scan clean.
```

A scanner that clears an obvious key is worse than no scanner, because it is trusted. #418 has a self-contained reproduction using the unmodified hook.

The noisier symptom is the same root cause: on PR #416 the hook blocked a push over an `api.pushover.net` URL that was already on `origin/main` and that the branch never touched, reported with a mangled `+ ++ +` gutter (the combined-diff column layout) and a meaningless line number. That trains people to reach for `--no-verify`, which skips the entire gate — ruff, bandit, OpenAPI, and the test suite included.

## What changed

`git log -p` accepts rev-list arguments natively and emits per-commit patches, which is the intended meaning, so both `git diff` calls become `git log`. Merge commits are skipped by default, which is right here: they carry nothing the individual commits in the range lack, and those commits are themselves in range.

Also added a backstop that fails closed when files were touched but no patch text came back at all, instead of reporting a clean scan. Its reach is deliberately documented as limited — the bad range still emitted combined-diff *headers*, so this check would **not** have caught the original bug. I verified that rather than assuming it. The tests are the real guard. It keys on files touched rather than on the commit count so an empty commit (`git commit --allow-empty`) stays pushable, which I confirmed was otherwise a spurious block I'd introduced.

## Tests

`tests/unit/test_prepush_hook.py` — the first coverage this hook has had. 9 tests drive the real script through its stdin protocol against throwaway repositories: new-branch and existing-branch pushes, a secret in a non-tip commit, `.env` additions, merge-only and empty-commit pushes, and branch deletion. Restoring the `git diff` call fails three of them.

They assert on scan *output* rather than exit status, because the hook's later ruff/bandit/pytest gate exits non-zero in a throwaway repo for unrelated reasons — exit status alone can't distinguish "the scanner blocked this" from "the gate failed".

Two things worth flagging, both found by running this rather than reasoning about it:

**The tests scrub `GIT_*` from the environment of every git call.** Without that they fail when run from inside the hook, which is exactly where the gate runs them: git exports `GIT_DIR` and `GIT_INDEX_FILE` to hooks, and inherited they point each fixture command at the real repository instead of the throwaway one. This is not theoretical — a trial run renamed my checked-out branch to `main`, clobbered the local `main` ref, and committed fixture state onto it. I repaired it from the reflog and then verified the scrubbed version leaves the host repo byte-identical (branch, HEAD, and `main` all unchanged) when run with `GIT_DIR` set.

**The fixtures assemble their secret-shaped strings at runtime** and render assignments through a helper, so the file contains no literal that this hook — or GitHub push protection — would flag, and no allowlist entry is needed. The fixed hook caught an earlier version of this file, which is a decent sign it works.

## Verification

Pushed with the fixed hook active and the full gate running: `pre-push: all checks passed`, 1523 passed / 2 skipped / 39 xfailed, no bypass. Compare with #416, which had to be pushed with `--no-verify` because of the false positive.

Note the hook is also verified under a real `git push`, not just standalone — the two differ, and that mattered. `core.hooksPath` here is an **absolute** path (`/Users/.../hashview/.githooks`), so pushes from a git worktree run the *main checkout's* hook, not the worktree's. My first attempts were silently exercising the old hook. Instrumenting a real push through the fixed one confirmed it sees the full diff (275 added lines, all three finding categories reported). That footgun is now addressed in the second commit, which documents next to the enable instruction that the value must stay relative. Git resolves a relative `hooksPath` against the top level of whichever working tree you run git from, so each worktree correctly uses its own `.githooks/` — verified from a repo root, a subdirectory, and a worktree. I also corrected my own clone's config, which was absolute; this PR's final push ran with no `-c core.hooksPath` override and correctly used the worktree's own hook (`pre-push: all checks passed`, 1523 passed).

Note this only fixes the documentation and my clone — `core.hooksPath` is per-clone local config, so anyone whose clone already has an absolute value needs to re-run `git config core.hooksPath .githooks` themselves. Checking with `git config core.hooksPath` is called out in the header.

No CHANGELOG entry: `.githooks/` changes are developer tooling and aren't changelogged here, matching the existing `chore(githooks):` commits.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -9,6 +9,14 @@
 # Enable once per clone:
 #   git config core.hooksPath .githooks
 #
+# Keep that value relative, exactly as written above. Git resolves a relative
+# hooksPath against the top level of whichever working tree you run git from, so
+# each git worktree correctly uses its own .githooks/. An absolute path (e.g.
+# /Users/you/projects/hashview/.githooks) instead pins every worktree to the
+# main checkout's copy: edits to this script made in a worktree then have no
+# effect on a push from it, and you are silently testing the old version.
+# Check with: git config core.hooksPath
+#
 # Bypass (use sparingly):
 #   git push --no-verify
 
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -69,18 +69,45 @@ while read -r local_ref local_sha remote_ref remote_sha; do
     commits=$(git rev-list "${range_args[@]}" 2>/dev/null)
     [ -z "$commits" ] && continue
 
-    # Combined diff of added lines only for the push range.
-    diff=$(git diff --unified=0 --no-color "${range_args[@]}" 2>/dev/null)
+    # Per-commit patch for the push range.
+    #
+    # Must be `git log -p`, not `git diff`: range_args uses rev-list syntax,
+    # and for a new branch that is `<sha> --not --remotes`. `git diff` does not
+    # understand --not/--remotes as range exclusion — it treats the extra revs
+    # as more sides to compare and emits a *combined* diff, which both omits
+    # real additions and attributes untouched upstream lines to the push. That
+    # made the scanner report "clean" on a commit holding an AWS key (#418).
+    # `git log -p` takes rev-list arguments natively. Merge commits are skipped
+    # by default, which is right: they add nothing the listed commits lack.
+    diff=$(git log -p --unified=0 --no-color --format='' "${range_args[@]}")
 
     # Files added/modified in this push.
-    changed_files=$(git diff --name-only --diff-filter=AM "${range_args[@]}" 2>/dev/null)
+    changed_files=$(git log --name-only --diff-filter=AM --format='' \
+                        "${range_args[@]}" | grep -v '^$' | sort -u)
+
+    # Every path touched by the push, any status. Used only for the sanity
+    # check below.
+    touched_files=$(git log --name-only --format='' "${range_args[@]}" \
+                        | grep -v '^$' | sort -u)
 
     # Commit messages (subject + body) for every commit in the push range,
     # prefixed with the short SHA so findings point at the offending commit.
     messages=$(git log --format='%h %B%n---END---' "${range_args[@]}" 2>/dev/null)
 
-    if [ -z "$diff" ] && [ -z "$messages" ]; then
-        continue
+    # Sanity check for gross breakage: files were touched but git returned no
+    # patch text at all. A backstop only — it catches a range that yields
+    # nothing, not one that yields the wrong thing. (In #418 the bad range
+    # still emitted combined-diff headers, so this check would NOT have caught
+    # it; the tests in tests/unit/test_prepush_hook.py are what guard that.)
+    #
+    # Keyed on touched_files rather than on the commit list, because an empty
+    # commit (git commit --allow-empty) legitimately has no patch and must
+    # still be pushable.
+    if [ -z "$diff" ] && [ -n "$touched_files" ]; then
+        printf '%spre-push: %s file(s) changed in this push but no diff could be computed.%s\n' \
+            "$RED" "$(printf '%s\n' "$touched_files" | wc -l | tr -d ' ')" "$RST" >&2
+        printf 'This is a bug in the hook, not a finding. Refusing to report a clean scan.\n' >&2
+        exit 1
     fi
 
     # --- Built-in patterns (scan added diff lines + commit messages) ---
--- a/tests/unit/test_prepush_hook.py
+++ b/tests/unit/test_prepush_hook.py
@@ -1,0 +1,277 @@
+"""The pre-push secret scanner must see the lines a push actually adds.
+
+Regression tests for #418. `.githooks/pre-push` built its scan range with
+`git rev-list` syntax (`<sha> --not --remotes`) and then handed those same
+arguments to `git diff`, which has no notion of `--not`/`--remotes` as range
+exclusion and emits a *combined* diff instead. On a new-branch push the
+scanner could therefore receive an empty haystack and report
+"sensitive-content scan clean." for a commit containing an AWS key.
+
+These tests drive the real hook script through its stdin protocol against
+throwaway repositories. They assert on the hook's scan *output* rather than
+its exit status: the hook also runs a ruff/bandit/pytest gate after the scan,
+which exits non-zero in a throwaway repo for unrelated reasons, so exit
+status alone cannot distinguish "blocked by the scanner" from "gate failed".
+"""
+
+import os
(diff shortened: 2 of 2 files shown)


bandrel:
CI note: `e2e-crack` failed on `5235b3b` and passed on re-run (both jobs, no code change in between). It also passed on the previous commit `c1f4a09`, and the entire diff between those two commits is 8 comment lines in `.githooks/pre-push` — a file the dockerized crack harness never reads, so it can't be causal. The failure was an app-side `HTTP 500` from `GET /v1/getHashType/1` during the agent's result upload, which looks like a flake in that suite rather than anything from this branch. Flagging it rather than quietly re-running, in case the 500 is worth chasing separately.

PR #432, opened 2026-08-26
fix: regenerate dynamic wordlists on the UI download

bandrel:
Stacked on #431 — please merge that first, or merge this into it.

## Problem

`GET /wordlists/download/<id>` serves whatever sits at `wordlist.path`. For a dynamic wordlist nothing keeps that file current: it is created as a zero-byte placeholder at install time (`add_default_dynamic_wordlists` opens it with mode `'w'`), and the only writer is the manual refresh button. So on a fresh install the operator downloads an empty file with a `200` and no indication anything is wrong.

Reproduced against a real container on `v0.8.3-dev`, with 10 recovered plaintexts in the database:

```
$ curl -b session /wordlists/download/1 -D -
HTTP/1.1 200 OK
Content-Disposition: attachment; filename=DYNAMIC_All_Recovered_Passwords.txt
Content-Length: 0
```

\#431 makes this strictly worse: it removes `/v1/updateWordlist/<id>` and has the API download generate into a per-request temp file, deliberately leaving the canonical path alone. After it merges, nothing writes `wordlist.path` during normal operation at all, so the UI download would serve the zero-byte placeholder forever.

## Change

Regenerate the list for the download, mirroring what #431 did for `GET /v1/wordlists/<id>`:

- Generation goes to a **per-request unique temp file**, never the shared `wordlist.path`, so concurrent downloads can't truncate the file out from under each other. The row's `size`/`checksum` metadata is deliberately left untouched.
- A dynamic list is derived from the database, so a missing on-disk file is no longer an error — it regenerates instead of redirecting.
- Static wordlists are unchanged: still served as the stored `.gz`, still redirect with a flash when the file is missing.

`_remove_file` / `_send_generated_file` were private to the API blueprint; they move to `hashview.utils.utils` as `remove_file` / `send_generated_file` so both blueprints share one copy of the `direct_passthrough` cleanup workaround (#226) rather than the UI growing a second one. That is a separate, behaviour-preserving commit.

## Tests

Six new tests in `tests/unit/test_downloads.py`, each watched failing first: regenerates from the database, never serves a stale file, works when the file is missing, leaves the canonical file and row metadata untouched, cleans up its temp files, and names the attachment after the wordlist rather than the random temp file.

Two existing tests pinned the old behaviour and were updated rather than deleted: the UI dynamic-download encoding round-trips now stub regeneration to reproduce the stored bytes (exactly as the `/v1` ones in #431 already do), and the rules-download gzip test patches `send_from_directory` where `send_generated_file` now resolves it.

## Verification

- `pytest tests/unit tests/security tests/agent_unit` — 1600 passed, 2 skipped, 40 xfailed, 5 xpassed
- `ruff check hashview/ hashview.py` — clean
- Docker stack built from this branch, against the same seeded database: the UI download of the length buckets and the All Usernames / All Recovered Passwords lists all return correct regenerated content while the canonical file stays 0 bytes; 5 consecutive downloads leave `control/tmp` unchanged; the API download and the static `.gz` download are unaffected.


bandrel:
Reopening to trigger CI.

Issue #422, opened 2026-08-21
Job-creation wizard does work proportional to whole tables, not to what each page shows (16 MB cracked-hash page, per-hashfile N+1, per-task form dump)


bandrel:
The job-creation wizard is fast on a small install and degrades as data accumulates, because four of its steps do work proportional to a whole table rather than to what the page actually shows. None of these are slow enough to notice on a fresh install, which is why they have survived; they become painful exactly when a customer has accumulated real engagement data.

Filing these together because they share a root cause and a fix strategy — render/query what the page displays, not the whole table — and because the wizard's total cost is what a user experiences, not any single step's.

## Measurements

Against the local docker stack (`docker-compose.yml`, MySQL 8) seeded to 150 hashfiles for one customer, 350k hashes, and 2,000 tasks. Medians of 5 samples, warmup discarded:

| Step | Median | Response size |
| --- | --- | --- |
| `GET /jobs/add` (per-request floor) | 237 ms | small |
| `GET /jobs/<id>/assigned_hashfile/` | 762 ms | 318 KB |
| `GET /jobs/<id>/tasks` | 324 ms | **1.5 MB** |
| `GET /jobs/<id>/assigned_hashfile/<id>` | **2,864 ms** | **16 MB** |
| `POST /jobs/<id>/assign_task/<id>` | 625 ms | — |
| `GET /jobs/<id>/summary` | 288 ms | 998 KB |

`GET /jobs/add` renders a two-field form, so its 237 ms is the fixed per-request overhead every other row also pays. Subtract it to see each step's own work.

## 1. Cracked-hash view loads and renders every cracked hash

`hashview/jobs/routes.py:465` loads the full result set as `(Hashes, HashfileHashes)` ORM objects with no pagination, and line 466 re-runs the same query as a `COUNT` rather than taking `len()` of what was just fetched:

```python
cracked_hashfiles_hashes = db.session.query(Hashes, HashfileHashes).join(...).filter(Hashes.cracked == '1').filter(HashfileHashes.hashfile_id==hashfile.id).all()
cracked_hashfiles_hashes_cnt = db.session.query(Hashes).join(...).filter(Hashes.cracked == '1').filter(HashfileHashes.hashfile_id==hashfile.id).count()
```

[...]

PR #434, opened 2026-08-26
fix(api): one grouped query for the hash-type hashfile listing (#228)

### LINKED ISSUE #228: API: N+1 queries in /v1/hashfiles/hash_type and /v1/jobs/add
## Summary
- `/v1/hashfiles/hash_type/<id>` (`hashview/api/routes.py:1162`) runs 2 count queries per matching hashfile inside a loop.
- `/v1/jobs/add` (`routes.py:766`) re-queries existing job tasks and commits inside the per-task loop.

## Suggested fix
Aggregate counts with a single grouped query; build task rows in memory and commit once outside the loop.

bandrel:
Fixes the N+1 in `GET /v1/hashfiles/hash_type/<t>` from #228.

> **Stacked on #433.** Base is `test/api-hashfile-listing-perf`, so the diff here is just the fix. GitHub retargets this to `v0.8.3-dev` automatically once #433 merges. Review #433 first.

## The change

The route took a `DISTINCT` list of matching hashfile ids and then, for each one, ran an ORM `Hashfiles.query.get()` plus two separate `COUNT`s — three round trips per matching file. Joining from `Hashfiles` and grouping does the same work in one statement.

Two things fall out of the rewrite:

- **The `if hashfile is None: continue` guard is gone**, subsumed by joining *from* `Hashfiles`. An orphaned junction row has nothing to join to, so it drops out rather than needing a lookup to discover it is an orphan. `test_hashfiles_by_hash_type_hashfile_missing_in_loop` is unchanged because the behaviour is; only its docstring moves off the deleted line reference.
- **Results are ordered by hashfile id.** `GROUP BY` does not imply an ordering on MySQL, and the old code iterated a `DISTINCT` list in whatever order came back, so this is newly deterministic.

## Measurements

Perf fixture: 31 hashfiles, 110k hashes, MySQL 8 in Docker. Loop cost is measured above the same-route zero-hashfile floor, per #433.

| | median | loop / floor | per hashfile |
| --- | --- | --- | --- |
| before | 1001 ms | 2.04x – 2.65x | 23 – 29 ms |
| after | **427 ms** | **0.49x** | **4.6 ms** |

For scale, on a production instance holding 7.85M NTLM hashes this endpoint took **549 seconds to return 13 KB** — past every default client timeout. Callers did not experience a slow listing, they experienced *no* listing, indistinguishable from a customer having no hashfiles of that type. That is why this is worth more than the 2.3x the local fixture shows.

## Output equivalence

Verified **byte-identical against MySQL**, not just against the sqlite unit fixtures — old and new code served from the same running stack and the same seeded data, compared for three cases:

- a populated type (31 hashfiles with real total/cracked counts) — identical
- an empty type (`22000`) — identical
- a type no hash has ever had (`4242`) — identical

## Tests

Two unit tests added for properties the rewrite makes load-bearing:

- **one row per hashfile regardless of hash count** — mutation-verified: removing the `GROUP BY` fails it (along with two existing tests)
- **`total == cracked` for a fully cracked file** — covers the all-ones end of `COALESCE(SUM(CASE ...))`, where the existing mixed-type test covers the zero end

The id-order assertion is deliberately called out in its docstring as **not** mutation-provable: sqlite returns grouped rows in rowid order regardless, so removing the `order_by` does not fail it. It documents contract, not a guard, and I would rather say so than let it read as coverage it is not.

The strict `xfail` from #433 is dropped. A marker being added, then clearing when the defect goes, is the workflow it was added for.

## Verification

- `pytest -m perf` (fix in place, `HASHVIEW_PERF_TASKS=2000`): **9 passed, 4 xfailed, 0 failed** — the four remaining xfails are #422's
- `pytest tests/unit tests/security tests/agent_unit`: **1622 passed, 2 skipped, 44 xfailed, 5 xpassed** (1620 before; the two new tests are mine, and the 5 xpassed are unchanged from baseline)
- `ruff check hashview/ hashview.py`: clean
- `pylint` at CI's exact invocation: **9.63/10, rc=0**, unchanged from baseline and above `fail-under=9.0`

## Scope

**This does not close #228.** That issue also covers `/v1/jobs/add` re-querying and committing inside a per-task loop. It is a write path and wants its own change with its own testing.

Two sibling call sites of the same loop are also left alone, deliberately:

- `hashview/hashfiles/routes.py:71` — the web UI hashfiles list, which runs it over *every* hashfile server-wide and has no issue of its own yet
- `hashview/jobs/routes.py:296` — the job-creation picker, which is #422's item 2

`/v1/customers/<id>/hashfiles` needs nothing: it runs one combined aggregate per hashfile rather than three queries, measures ~6 ms per hashfile, and is already under #433's ceiling. See my measurements on #228.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ Notable changes will be documented here
 
 **Testing**
 - Job-creation performance e2e suite (`tests/e2e/test_job_creation_perf.py`) with a tunable volume seeder (`tests/seed_perf_db.py`), covering latency budgets for each wizard step plus strict-xfail scaling guards for the four endpoints in issue #422 whose cost grows with table size rather than page size
-- API hashfile-listing performance suite (`tests/e2e/test_api_hashfile_listing_perf.py`), sharing that seeder. Asserts loop cost as a ratio to a same-route zero-hashfile floor rather than an absolute budget, so the threshold survives a change of hardware; carries a strict-xfail scaling guard for issue #228 (`GET /v1/hashfiles/hash_type/<t>` runs an ORM get plus two counts per matching hashfile) and a passing regression guard on `GET /v1/customers/<id>/hashfiles`, which runs one combined aggregate per hashfile and measures under the ceiling today
+- API hashfile-listing performance suite (`tests/e2e/test_api_hashfile_listing_perf.py`), sharing that seeder. Asserts loop cost as a ratio to a same-route zero-hashfile floor rather than an absolute budget, so the threshold survives a change of hardware; carries regression guards for issue #228 (fixed in this release; the guard remains so a per-hashfile loop cannot return) and on `GET /v1/customers/<id>/hashfiles`, which runs one combined aggregate per hashfile and measures under the ceiling today
 
 ### Changed
 - The Rules listing is paginated (20 per page) with sortable Name / Rules / Owner / Last Updated columns and a server-side name filter, matching the Tasks listing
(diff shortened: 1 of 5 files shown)


i128:
Looks good thanks

Issue #314, opened 2026-07-08
max file size for pasted hashes


i128:
Werkzeug's max_form_memory_size is set to 488KB. This means any/all forms in the app are limited to entries under this size.
File uploads are not limited this way. If enough demand is made for it, we can add some sort of setting to deal with it. But for now will leave as an open issue.

bandrel:
Correcting one detail while addressing this: `max_form_memory_size` isn't a Hashview setting — it appears nowhere in this repo or its git history. The 488KB figure is Flask 3.1's own built-in `MAX_FORM_MEMORY_SIZE` default (500000 bytes), inherited transitively since this repo doesn't pin Werkzeug.

Worth noting as a workaround in the meantime: the `/v1` API's paste endpoint (`hashview/api/routes.py:1528`) reads the raw request body via `request.get_data(as_text=True)` rather than form-parsing, so it isn't subject to this cap at all.

Opened #457, which makes the cap configurable (`MAX_FORM_MEMORY_SIZE` under `[SERVER]` in `config.conf`, defaulting to Flask's own 500000 so nothing changes unless a deployer opts in) and adds the app's first `@app.errorhandler(413)` — previously exceeding the limit produced a bare Werkzeug error page or a misleading "You must assign a name to the hashfile" flash instead of anything pointing at the real cause/workaround.

PR #457, opened 2026-09-04
feat: make the pasted-hash form-size cap configurable and add a 413 handler

bandrel:
## Summary
Issue #314's premise was slightly off: `max_form_memory_size` (~488KB) isn't a Hashview setting — it's Flask 3.1's own built-in `MAX_FORM_MEMORY_SIZE` default (500000 bytes), which appears nowhere in this repo or its history. This PR:

- Adds a `MAX_FORM_MEMORY_SIZE` key under `[SERVER]` in `config.conf`/`config.conf.example`, read via the same `.get(key, default)` idiom already used for `SECRET_KEY` (not the bare-indexing idiom used for `SERVER_NAME`, which would `KeyError` on an older config.conf). Defaults to 500000 — Flask's own default — so leaving it unset changes nothing. Note: `config.conf` is baked into the Docker image at build time, so this must be set before `docker compose build`.
- Registers the app's first-ever `@app.errorhandler(413)`. Previously, exceeding the limit produced a bare Werkzeug 413 page or a misleading "You must assign a name to the hashfile" flash. Now it matches the existing AJAX/normal-submit contract used elsewhere (`X-Requested-With: fetch` → JSON `{status, msg}`; normal submit → flash + redirect), with a message pointing at the real workaround: upload the hashes as a file instead, since file uploads aren't subject to this cap.

Closes #314

## Test plan
- [x] New tests trigger a real 413 (tiny configured limit + oversized POST, not a mocked exception): JSON shape for AJAX, flash+redirect for normal submit
- [x] Test confirms `Config.MAX_FORM_MEMORY_SIZE == 500000` when the key is absent, matching Flask's own default
- [x] Full unit suite: 1539 passed, 2 skipped, 46 xfailed, 1 xpassed (pre-existing, unrelated flake)
- [x] ruff clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,10 @@ Notable changes will be documented here
 - Bulk-select and bulk-delete on the hashfiles list
 - Per-hashfile `--hex-salt` option for salted hash types
 
+**Configuration**
+- Optional `MAX_FORM_MEMORY_SIZE` key under `[SERVER]` in `config.conf`, capping the size of pasted (non-file) form data such as the pasted-hashes textarea. Defaults to Flask's own 500000 bytes when unset, so existing deployments are unaffected; raise it to allow larger pastes. Uploaded files are not counted against it, and values below 65536 are clamped up to that floor because a smaller cap would reject multipart file uploads outright (#314)
+- Oversized submissions now return a clear "try uploading the hashes as a file instead" message — JSON for AJAX requests, a flash and redirect for normal form posts — rather than a bare Werkzeug 413 page or a misleading "You must assign a name to the hashfile" error (#314)
+
 **Testing**
 - Job-creation performance e2e suite (`tests/e2e/test_job_creation_perf.py`) with a tunable volume seeder (`tests/seed_perf_db.py`), covering latency budgets for each wizard step plus strict-xfail scaling guards for the four endpoints in issue #422 whose cost grows with table size rather than page size
 - API hashfile-listing performance suite (`tests/e2e/test_api_hashfile_listing_perf.py`), sharing that seeder. Asserts loop cost as a ratio to a same-route zero-hashfile floor rather than an absolute budget, so the threshold survives a change of hardware; carries regression guards for issue #228 (fixed in this release; the guard remains so a per-hashfile loop cannot return) and on `GET /v1/customers/<id>/hashfiles`, which runs one combined aggregate per hashfile and measures under the ceiling today
--- a/hashview/__init__.py
+++ b/hashview/__init__.py
@@ -2,7 +2,7 @@ import datetime
 import logging
 from logging.config import dictConfig as loggingDictConfig
 
-from flask import Flask, redirect, request, url_for
+from flask import Flask, flash, jsonify, redirect, request, url_for
 from jinja2 import select_autoescape
 
 __version__ = '0.8.3'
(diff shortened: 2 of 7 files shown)


i128:
Does this actually get enforced on form submission, specifically when user submits large hash files? i dont see how. Also if its set to a number larger than 500k, does flask accept it?

bandrel:
Good questions — I tested both against the pinned versions (Flask 3.1.3 / Werkzeug 3.1.8) rather than going off the docs, and the second one turned up a real bug in this PR.

### "Does this actually get enforced on form submission? I don't see how."

It is enforced, but nothing in Hashview does the enforcing — that's why it isn't visible in the diff. Werkzeug applies it inside the form parser while reading the request body, at [`werkzeug/formparser.py:383-389`](https://github.com/pallets/werkzeug/blob/3.1.8/src/werkzeug/formparser.py#L383-L389):

```python
elif isinstance(event, Data):
    if self.max_form_memory_size is not None and field_size is not None:
        field_size += len(event.data)
        if field_size > self.max_form_memory_size:
            raise RequestEntityTooLarge()
```

That runs before any view function is dispatched, which is why the 413 has to be caught with a top-level `errorhandler` instead of a check inside a route.

### "...specifically when a user submits large hash files?"

It does **not** cap uploaded files, and you're right to be suspicious of that framing. Note `field_size is not None` in the snippet above — `field_size` is initialized to `None` for `File` events and only to `0` for `Field` events (formparser.py:374-379), so file parts stream to disk uncounted. The total-body cap that *would* limit them, `MAX_CONTENT_LENGTH`, defaults to `None`.

Confirmed empirically: a 50 MB file upload succeeds at the default `MAX_FORM_MEMORY_SIZE` of 500000.

So this setting only affects the pasted-textarea path, which is what the PR intends.

### "If it's set to a number larger than 500k, does Flask accept it?"

Yes, with no secondary ceiling. A 2 MB pasted field with the cap at 5000000 returns 200; the same field at the 500000 default returns 413.

### The bug this shakes out

My config comment says "File uploads are NOT subject to this limit." **That is wrong for small values**, and I should fix it before this merges.

There's a *second* enforcement point I missed, in the multipart decoder at [`werkzeug/sansio/multipart.py:130-136`](https://github.com/pallets/werkzeug/blob/3.1.8/src/werkzeug/sansio/multipart.py#L130-L136):

```python
elif (
    self.max_form_memory_size is not None
    and len(self.buffer) + len(data) > self.max_form_memory_size
):
    raise RequestEntityTooLarge()
```

This one is chunk-level and does *not* distinguish file parts from field parts. The parser reads in `buffer_size` chunks, which defaults to `64 * 1024` (formparser.py:298), so **any value below 65536 breaks file uploads entirely**:

| `MAX_FORM_MEMORY_SIZE` | 10 KB **file** upload |
| --- | --- |
| 1000 | 413 |
| 65536 | 200 |
| 500000 | 200 |

The failure mode is nasty: a deployer who *lowers* this to tighten things up silently loses all file uploads — including the exact workaround the new 413 message tells them to use. And my own test fixture uses `MAX_FORM_MEMORY_SIZE: 16`, which sits inside that broken regime, so the suite never exercises a realistic config.

Fix I'll push: clamp the configured value to a floor of 65536 in `hashview/config.py`, correct the `config.conf.example` comment to state the real constraint, and add a test asserting a file upload still succeeds at the minimum. Holding the merge until that's in.

bandrel:
Fixes pushed — 37d2908 and e942a6a. Ready for another look, @i128.

- **Floor + clamp.** Values below 65536 are clamped up to it. Resolution moved into a new dependency-free `hashview/form_limits.py`, since `hashview/config.py` reads `config.conf` at import time and so can't be imported from the test suite. A typo in the key now falls back to the default rather than raising at import.
- **Corrected the `config.conf.example` comment**, which flatly claimed file uploads were exempt from the limit. It now states the actual constraint.
- **Replaced the `MAX_FORM_MEMORY_SIZE: 16` test fixture** with the real floor and genuinely oversized bodies, plus a new test asserting a 1MB file upload still parses at the minimum cap — i.e. that the workaround the 413 message recommends actually works at the tightest setting an operator can configure.
- CHANGELOG entry added.

### One correction to my previous comment

I said a sub-64KiB cap breaks file uploads outright. Writing the test showed that's too strong, and I'd rather flag it than leave a wrong claim in the thread.

The chunk check is `len(buffer) + len(data) > max_form_memory_size`, and each read is at most `buffer_size`. So below the floor an upload fails only once the body outgrows the cap — a 1000-byte cap rejects a 10KB hash file, but a 500-byte one still goes through:

| cap | 5KB file | 10KB file | 200KB file |
| --- | --- | --- | --- |
| 16 | 413 | 413 | 413 |
| 1000 | 200 | 413 | 413 |
| 65535 | 200 | 200 | 413 |
| 65536 | 200 | 200 | 200 |

The floor of 65536 is still the right number, and for the reason that matters: at or above it *no single read can ever exceed the cap*, so uploads parse cleanly at any file size. Below it the breakage is size-dependent, which is arguably worse to debug in the field — uploads work in testing with a small file, then fail on a real hashfile. Both behaviours are pinned by parametrized tests in `tests/unit/test_form_memory_limit_floor.py`.

Also confirmed while in there: `MAX_CONTENT_LENGTH` is still `None`, so there's no total-body cap on hashfile uploads. That's the pre-existing behaviour and I haven't touched it, but if you'd want an upper bound on upload size that's a separate knob and worth its own issue.

Mutation-checked the clamp (removing it fails 6 tests). Full unit suite: 1706 passed, 2 skipped, 61 xfailed, 1 xpassed (pre-existing).

i128:
THanks!

Issue #439, opened 2026-08-28
TaskGroups.name has no unique DB constraint (TOCTOU on duplicate-name check)


bandrel:
## Summary

`TaskGroups.name` (`hashview/models.py`) has no `unique=True` and no unique
index at the DB level. Both the UI (`TaskGroupsForm.validate_name` in
`hashview/task_groups/forms.py`) and the new `/v1` API
(`POST /v1/task_groups/add`, added in #401) enforce uniqueness with a
check-then-act query: `TaskGroups.query.filter_by(name=name).first()` before
`db.session.add(...)`.

## Impact

Two concurrent creates with the same name can both pass the check and both
commit, leaving two `TaskGroups` rows sharing a name. This was a pre-existing
gap in the UI form; adding a scriptable `/v1` endpoint makes concurrent
creates far more likely than a human clicking a "New group" modal, so it's
worth closing now.

Failure mode is duplicate rows, not data corruption — no cascading breakage,
but it is confusing (`GET /v1/task_groups` would list two groups with the
same name and no way for a client to tell them apart except id).

## Proposed fix

A migration adding a unique index on `task_groups.name`, plus catching the
resulting `IntegrityError` in both the UI route and the API route to
surface the existing "name already taken" message instead of a raw 500.

Found while reviewing hashview-utils' bulk task tooling and implementing
#401.

PR #460, opened 2026-09-06
fix: unique constraint on task_groups.name to close a TOCTOU race

bandrel:
## Summary
`TaskGroups.name` had no unique DB constraint — only a check-then-act query in both the UI form and the `/v1` API — so two concurrent creates with the same name could both pass validation and both commit.

- `hashview/models.py` — named unique constraint `uq_task_groups_name` on `TaskGroups.name`.
- New migration `b5c8d9e1f2a4` (chained onto head `d3a4a6a7b352`), guarded and dialect-neutral. It scans for pre-existing duplicate names before applying and raises a clear error naming them if found (MySQL DDL is non-transactional, so a blind `ADD CONSTRAINT` would strand mid-apply on a real deployment). Uses `op.batch_alter_table` — plain `create_unique_constraint` raises `NotImplementedError` on SQLite outside batch mode, which would have made the real DDL line untestable and untested outside a live MySQL run. Verified by forcing the actual downgrade/upgrade DDL cycle against SQLite.
- `DEV_HEAD` bumped to `b5c8d9e1f2a4` in `tests/run_migration_e2e.sh`, `tests/unit/test_migration_drift_idempotency.py`, `tests/integration/test_migration_e2e.py`.
- `hashview/task_groups/routes.py` — `task_groups_add` (both the modal flow and the legacy standalone page) and `task_groups_edit` now catch `IntegrityError`, roll back, and surface "That task group name is taken." The legacy page needed its own fix: the template only renders `field.errors`, not an arbitrary message kwarg.
- `hashview/api/routes.py` — `v1_api_add_task_group` catches `IntegrityError` before the generic handler and returns the same 400 message the existing pre-check already returns (`{'status': 400, 'msg': 'A task group with that name already exists'}`), with a rollback the old generic handler was missing.
- New tests force real races (bypassing the app-level pre-check) for all four write paths: API, modal create, legacy create, edit. Verified via coverage that each actually executes its `except IntegrityError` branch, not just the ordinary pre-check.

Closes #439

## Test plan
- [x] ruff, bandit (vs baseline), openapi-spec-validator, `pytest tests/unit tests/security tests/agent_unit`: 1729 passed, 2 skipped, 61 xfailed
- [x] Manual SQLite check: constraint enforced (duplicate insert → IntegrityError), guarded upgrade is idempotent (second upgrade no-ops), and the real batch-mode DDL (not just the guard) verified via forced downgrade→upgrade cycle
- [ ] `tests/run_migration_e2e.sh` (real MySQL) not run in this environment — a pre-existing unrelated container was occupying port 3306. Recommend running it before merge.

### DIFF
--- a/hashview/api/routes.py
+++ b/hashview/api/routes.py
@@ -13,6 +13,7 @@ from flask import (
 )
 from packaging import version
 from sqlalchemy import case, exists, func
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.declarative import DeclarativeMeta
 
 import hashview
@@ -1370,7 +1371,15 @@ def v1_api_add_task_group():
         task_group = TaskGroups(name=name, owner_id=user.id, tasks=json.dumps(ordered))
         db.session.add(task_group)
         db.session.commit()
+    except IntegrityError:
+        db.session.rollback()
+        return jsonify({
+            'status': 400,
+            'type': 'Error',
+            'msg': 'A task group with that name already exists'
+        })
     except Exception:
+        db.session.rollback()
         current_app.logger.exception('API /v1/task_groups: failed to add task group')
         return jsonify({
             'status': 500,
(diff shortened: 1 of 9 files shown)


i128:
Thank you

PR #466, opened 2026-09-06
perf(#422): paginate cracked-hash view, narrow columns, server-side filter

bandrel:
## Summary
Part of issue #422 (defect 1 of 4, and the worst absolute number: a 50k-hash file with 30k cracked produced a 16 MB page in ~2.9s). The cracked-hash view loaded and rendered every cracked hash in a hashfile with no pagination, plus a duplicate `COUNT` query solely for the flash message.

- `hashview/jobs/routes.py` — paginates via `.paginate(page=page, per_page=20, error_out=False)`, matching the `rules_list` convention. Narrows the query to `.with_entities(HashfileHashes.username, Hashes.ciphertext, Hashes.plaintext)` — verified via direct SQL inspection that only those three columns are selected (not full ORM row objects). Drops the duplicate `.count()` in favor of `pagination.total`.
- Added a server-side `?q=` text filter (searches plaintext/ciphertext/username, properly escaped for `%`/`_`/backslash), since the old client-side JS filter would have silently only searched the current page post-pagination — a "looks like it searches everything, actually doesn't" regression a user wouldn't notice.
- `hashview/templates/jobs_assigned_hashfiles_cracked.html.j2` — pagination controls added; client-side filter JS and `data-search` attributes removed.

**Two real bugs found by two independent review passes (both converged on the same first one) and fixed before this PR, verified with regression tests:**
1. **Missing `ORDER BY`** — `LIMIT`/`OFFSET` across this two-table join had no guaranteed row order. Harmless by accident on SQLite (rowid order), but MySQL/InnoDB makes no such guarantee, so rows could repeat or be skipped between pages as the query planner shifts. Added `.order_by(Hashes.id)`, matching the pattern already used at `jobs/routes.py:103`. **Honesty note**: I confirmed empirically that the new "pages partition the result set" test still passes even with this fix removed, on SQLite — this is a real MySQL-only correctness issue the unit-test suite structurally cannot observe. The fix is still correct and necessary for production.
2. **Empty-state message inverted** — the template branched on `pagination.total > 0` (the *filtered* count) to decide whether to show "no matches for your filter" vs. "no cracked hashes at all". A filter matching nothing always has `total == 0`, so it fell through to the false statement "No previously cracked hashes found" on a hashfile that actually has cracked hashes. Fixed to branch on `search_filter` directly, plus added a distinct branch for a manually-edited out-of-range `?page=`.

Also fixed as part of the same review pass: the flash message was showing the *filtered* count instead of the hashfile-wide total once a search was active (restored to match pre-change behavior), and factored the LIKE-escaping logic into a shared `escape_like_term()` helper in `hashview/utils/utils.py` (it was a byte-for-byte duplicate of logic inside `apply_name_filter`).

Closes #422 (defect 1)

## Test plan
- [x] ruff, bandit (vs baseline), openapi-spec-validator, `pytest tests/unit tests/security tests/agent_unit`: 1735 passed, 2 skipped, 61 xfailed
- [x] Verified via direct SQL inspection that the query selects exactly 3 columns + LIMIT/OFFSET, not full ORM objects
- [x] New tests cover: pagination bounds row count, second page has distinct non-overlapping rows, flash shows grand total (not page count, not filtered count), server-side filter narrows by plaintext/ciphertext/username, filter-matches-nothing empty state, pages partition the full result set with no gaps or repeats
- [x] Each new regression test verified via revert-and-confirm-it-fails against the pre-fix code
- [ ] `tests/e2e/test_job_creation_perf.py::test_cracked_hashes_payload_is_bounded` (the companion strict-xfail perf test) not run — needs a seeded live Playwright stack unavailable in this environment. Left the marker in place; whoever runs it against a live stack should remove it once confirmed.
- [ ] **Recommend running the real migration/MySQL-backed test suite (or at least a manual click-through) before merge**, specifically to confirm pagination behaves correctly under MySQL's join planner — the ORDER BY fix addresses a failure mode that cannot be observed on SQLite.

### DIFF
--- a/hashview/jobs/routes.py
+++ b/hashview/jobs/routes.py
@@ -15,7 +15,7 @@ from flask import (
 )
 from flask_login import current_user, login_required
 from flask_wtf import FlaskForm
-from sqlalchemy import case, func
+from sqlalchemy import case, func, or_
 
 from hashview.jobs.forms import (
     JobsForm,
@@ -45,6 +45,7 @@ from hashview.utils.utils import (
     apply_name_filter,
     build_job_task_commands,
     dynamic_wordlist_ids,
+    escape_like_term,
     import_hashfilehashes,
     save_file,
     try_commit,
@@ -493,8 +494,14 @@ def jobs_assigned_hashfile_cracked(job_id, hashfile_id):
     # Apply server-side search filter on plaintext, ciphertext, or username
     # Escape the search term once, then reuse for all columns
     if search_filter:
-        from sqlalchemy import or_
-        escaped = search_filter.replace('\\', '\\\\').replace('%', '\\%').replace('_', '\\_')
+        # Grand total across the whole hashfile, independent of the filter --
+        # the flash message reports "this hashfile has N instacracks" (as it
+        # did before pagination/filtering existed), not "N results for your
+        # search". pagination.total below is the FILTERED count and is wrong
+        # for this message once a filter is active.
+        hashfile_total_cracked = base_query.count()
+
+        escaped = escape_like_term(search_filter)
         base_query = base_query.filter(
             or_(
                 Hashes.plaintext.ilike('%' + escaped + '%', escape='\\'),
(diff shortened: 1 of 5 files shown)


i128:
I get that this page is slow, and that pagenation would speed things up, but i want the full list displayed on this step.

Issue #471, opened 2026-09-07
Customer-delete hash-cleanup check uses distinct('customer_id') on a model that has no such column


bandrel:
## Description

\`hashview/customers/routes.py\` around line 194-195, in the customer-delete flow:

\`\`\`python
for hash in hashes:
    # Check to see if our hashfile is the ONLY hashfile for this customer that has this hash
    customer_cnt = HashfileHashes.query.filter_by(hash_id=hash.id).distinct('customer_id').count()
    if customer_cnt < 2:
        db.session.delete(hash)
        ...
\`\`\`

\`HashfileHashes\` (hashview/models.py) has no \`customer_id\` column — only \`id\`, \`hash_id\`, \`username\`, \`hashfile_id\`. \`customer_id\` lives on \`Hashfiles\`, and this query never joins to it.

On non-PostgreSQL backends (this project runs MySQL/MariaDB in prod, SQLite in unit tests), SQLAlchemy's \`distinct(column)\` (i.e. \`DISTINCT ON\`) is silently ignored — confirmed by the \`SADeprecationWarning\` this line already emits in the unit test run: "DISTINCT ON is currently supported only by the PostgreSQL dialect... silently ignored". So in practice this just counts every \`HashfileHashes\` row referencing \`hash.id\`, across every hashfile/customer — not the number of distinct customers, and the \`customer_id\` distinct clause does nothing at all.

## Impact

The comment's stated intent — "only delete this shared \`Hashes\` row if no *other customer's* hashfile still references it" — is not what the code checks. It actually counts hashfile references to that hash regardless of customer, and even that count is affected by the ignored/no-op distinct clause. Depending on data shape, this can cause a \`Hashes\` row to be incorrectly kept or incorrectly deleted when a customer is removed (deleting a hash that's still referenced by another customer's hashfile, or vice versa, is the failure mode a correct dedup check should prevent).

## Suggested fix

Join to \`Hashfiles\` to scope by \`customer_id\` correctly, e.g. count distinct \`Hashfiles.customer_id\` values (or distinct \`hashfile_id\` if that's the intended unit) for hashfiles referencing \`hash.id\`, excluding the customer currently being deleted.

Found during a repo-wide inline-comment accuracy sweep (see PR docs/inline-comment-accuracy).

bandrel:
**Correction after empirical verification.** I tested this directly rather than relying on the static-analysis read above, and the impact claim needs to be walked back.

[...]

Issue #470, opened 2026-09-07
fig8 username==password download never decodes $HEX[...] passwords despite the comment


bandrel:
## Description

\`hashview/analytics/routes.py\` around line 774 (the fig8 username==password download route) has:

\`\`\`python
for entry in fig8_cracked_hashes:
    if entry[1] and entry[0]:
        # Decode username (handle possible domain delimiters)
        raw_username = entry[1]
        ...
        # Decode password
        password = entry[0]

        if username == password:
            fig8_usernames.append(username)
\`\`\`

The "Decode password" comment claims decoding happens, but \`password = entry[0]\` is a bare assignment with no decoding — unlike the main dashboard query (same file, ~line 338) which calls \`decode_hex_plain(plaintext)\`.

## Impact

If a recovered plaintext is stored as \`$HEX[...]\` (non-UTF-8 bytes), it will never equal the decoded \`username\`, so that username==password match is silently excluded from the fig8 export — even though the Shared Passwords / dashboard views would decode and potentially match it correctly.

## Suggested fix

Decode \`entry[0]\` the same way the main dashboard query does (\`decode_hex_plain\`) before the \`username == password\` comparison.

Found during a repo-wide inline-comment accuracy sweep (see PR docs/inline-comment-accuracy).

Issue #469, opened 2026-09-07
fig9 shared-password download never decodes $HEX[...] usernames despite the comment


bandrel:
## Description

\`hashview/analytics/routes.py\` around line 705 (the fig9 shared-passwords download route) has:

\`\`\`python
for entry in fig9_usernames:
    if entry:
        # Decode possible hex‑encoded usernames
        try:
            decoded = entry
        except Exception:
            decoded = entry
        outfile.write(f"{decoded}\n")
\`\`\`

The comment claims hex-decoding, but both the try and except branches just assign \`decoded = entry\` unchanged — this is a no-op. The main dashboard query path (line ~338 in the same file) correctly calls \`decode_hex_plain(plaintext)\` for the equivalent case, so this download path is inconsistent with it.

## Impact

A username stored as \`$HEX[...]\` (non-UTF-8 bytes) is exported from the fig9 download still wrapped in \`$HEX[...]\`, instead of the decoded form shown elsewhere in the UI/exports.

## Repro

1. Import a hashfile with a non-UTF-8 username so it's stored as \`$HEX[...]\`.
2. Get that username into the fig9 shared-passwords set.
3. Download the fig9 export — the username appears as \`$HEX[...]\` instead of decoded.

## Suggested fix

Call the existing \`decode_hex_plain\` helper (or equivalent) inside the try block instead of the no-op assignment, matching the main dashboard query's handling.

Found during a repo-wide inline-comment accuracy sweep (see PR docs/inline-comment-accuracy).

PR #473, opened 2026-09-07
test: strict-xfail the analytics hex-decode no-ops, drop a fixed xfail

bandrel:
## Summary
- Adds strict-xfail regression tests for #469 and #470, both found while auditing inline comments for accuracy in `hashview/analytics/routes.py`:
  - fig9 shared-hash download: the "Decode possible hex-encoded usernames" comment sits above a no-op (`decoded = entry`) — a `$HEX[...]`-wrapped username is exported still hex-wrapped.
  - fig8 username==password download: the "Decode password" comment sits above a bare assignment (`password = entry[0]`) — a `$HEX[...]`-wrapped recovered password never matches its username and is silently excluded.
  - Both were verified live against a real app before writing the xfail (not just inferred from reading the code).
- Drops the `xfail` marker on `test_customers_delete_with_uncracked_hash_succeeds`: it documented a TypeError bug (routes.py:185-186, a raw `Query` object compared to an int) that's already fixed by a `.count()` call. It's been silently XPASSing (non-strict) — now a plain passing regression test.
- A third finding at the same customer-delete line (`distinct('customer_id')` is a no-op since `HashfileHashes` has no such column, filed as #471) is **not** pinned with an xfail here — I verified two live scenarios (hash shared across customers; hash used only within one customer's multiple hashfiles) and both produced the *correct* outcome today, because SQLAlchemy's autoflush ordering happens to make the raw row-count converge to the right answer despite the no-op distinct clause. Filed as a fragile/misleading-code cleanup rather than a demonstrated bug (see issue comment on #471 for the full trace).

## Test plan
- [x] `ruff` clean
- [x] `bandit` vs. baseline clean
- [x] OpenAPI spec validates
- [x] Full unit/security/agent suite passing, xfail count net +1 (two new strict xfails, one dropped)


i128:
Please resolve merge conflict

PR #474, opened 2026-09-08
API: PUT /v1/rules/<id> and /v1/wordlists/<id> to replace content in place (#399)

bandrel:
## Summary

Closes #399. Adds a way to replace the *contents* of an existing rule or wordlist without deleting/recreating it — previously the only write routes were `POST /v1/rules/add/<name>` and `POST /v1/wordlists/add/<name>`, both of which always mint a new row, and delete is blocked precisely when a resource is wired into a task (the case that most needs refreshing).

- `PUT /v1/rules/<int:rules_id>` — overwrites the rule file in place (same id, same name, same path); body may be plain text or gzip (decompressed server-side, rules are plaintext at rest).
- `PUT /v1/wordlists/<int:wordlist_id>` — re-ingests the body via the existing `ingest_static_wordlist_file` helper, repoints the row (same id/name), and removes the old on-disk file; refuses `type: dynamic` wordlists with 400 (those are regenerated from the DB).
- Both refuse with **409** if a `Tasks` row using the resource has a currently `Running` `JobTasks` entry — swapping content out from under an in-flight crack would silently change what that run is doing. New shared helper: `resource_in_running_task(rule_id=..., wl_id=...)`.
- The existing web UI rule editor (`/rules/edit/<id>`) gets the same 409-equivalent guard — it previously had none.
- New shared helper `replace_file_atomic()` handles the `control/tmp` → live-path swap, tolerating `control/` being a separate mount point (`os.replace` raising `EXDEV`, see #395) via a same-directory copy + replace rather than a truncate-in-place copy.
- OpenAPI spec (`hashview/api_docs/openapi.yaml`) updated with both new operations; CHANGELOG updated.

## Test plan

- [x] `pytest tests/unit -q` — 1700 passed, 2 skipped, 61 xfailed, 1 xpassed (pre-existing, unrelated: a `Query` vs `int` comparison bug in `customers/routes.py`), 0 failed
- [x] `ruff check` clean on all touched files
- [x] New unit coverage: helper functions (`resource_in_running_task`, `replace_file_atomic` incl. EXDEV fallback), both PUT routes (success/404/403/409/400-empty-body/400-bad-gzip/400-dynamic/agent-rejected), and the web UI edit-guard regression test
- [x] `tests/unit/test_openapi_spec.py` — spec/route parity passes with the new PUT operations

### DIFF
--- a/.githooks/safe-allowlist.txt
+++ b/.githooks/safe-allowlist.txt
@@ -104,3 +104,9 @@ FCF7C1B8749CF99D88E5F34271D636178FB5D130
 # the shared prefix so the pattern also matches this very line, which the
 # generic secret-assign rule flags whenever it is added or moved.
 api_key="user-api-key-
+
+# --- tests/unit/test_api_endpoints.py (issue #399 PUT rules/wordlists) ---
+# Dummy api_key literals for the owner/non-owner ownership-check fixtures.
+# Not real secrets.
+api_key="owner-key"
+api_key="other-key"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Notable changes will be documented here
 - `POST /v1/task_groups/add` -- create a task group and its ordered task membership in one call (#401)
 - `POST /v1/task_groups/<task_group_id>/tasks` -- replace or append a task group's task membership (#401)
 - `DELETE /v1/task_groups/<task_group_id>` -- delete a task group, owner/admin only (#401)
+- `PUT /v1/rules/<id>` and `PUT /v1/wordlists/<id>` to replace an existing rule's or static wordlist's content in place (same id, same name) — refused with 409 if a task using the resource has a currently running job (#399)
 
 **Encrypted Database Backup**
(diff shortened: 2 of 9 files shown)


i128:
This should be updated to have the server send a 409 for HTTP PUT for rules associated with tasks that are also queued. 

Also impact to consider, if a rule set changes, then fundamentally the task has changed, as a result should the metrics around the effectiveness of that task be wiped?

If i have a wordlist like rockyou.txt + best64.rule, and i replace the contents of best.64 by deleteing all but one, then when i look at my task metrics, it'll show the effectiveness (recovery rate) of the task pre-edit.

I'd hate to loose the historical data, but if the rule got reset, then the recoverd_by value in hashes table should also be reset too. 

Then if thats done, what should those values be set to?

also

I dont agree that we should allow people to update existing static wordlists. I'd rather a new one be created.

bandrel:
You're probably right. The main reason that i was thinking was rule development only adds not removals, but you're right about stats and that would make it difficult to track. Closing as wont implement

PR #476, opened 2026-09-09
Fix/krb5tgs 18 spn parsing

i128:
## Fix Kerberos hash import: impacket SPN field, case-folded salts, and validator/hashcat parity

Reported symptom: kerberoasting with `GetUserSPNs.py` produces AES tickets that hashview rejects, e.g.

```
$krb5tgs$18$user$REALM$*MSSQLSvc/host.dom.local:1433*$368fe325e40917e65ba84d74$7a904516...
```

Users were hand-deleting the `$*...*` segment to get a hash accepted. Chasing that turned up two further
defects on the same path, one of them silent, so this branch fixes three things.

### 1. The reported bug — the SPN field was rejected (`bcc6307`)

`_KERBEROS_RE` allowed exactly four `$`-separated fields for 19600/19700 (user, realm, checksum, edata2),
with no slot for the SPN. All three import paths were affected: the wizard's paste box and file upload
(both reach `jobs/routes.py:375`, the paste via a temp file) and `POST /v1/hashfiles`
(`api/routes.py:1624`).

The field is now **optional**, not required — hashcat's own example has no SPN, and hashes already stored
via the manual workaround must keep validating. It is specifically *not* the 13100-style
`*user$realm$spn*` triple, which hashcat rejects for etype 17/18; Invoke-Kerberoast and PowerView emit
that shape and stay correctly blocked.

**The regex alone would have been a regression.** hashcat drops the SPN from its output, and a recovered
hash is matched to its row by an exact md5 of the stored ciphertext (`api/routes.py:2002`, no fallback
outside mode 22000). Storing the SPN would mean the job runs, hashcat recovers the password, and the
server silently discards it — worse than the loud rejection it replaced. Measured md5s for that pair
differ (`097f3c29…` stored vs `fb0238f6…` echoed). Kerberos imports therefore now store the shape hashcat
hands back. 13100 is untouched: its triple round-trips verbatim, which is why nobody hit this before.

### 2. Case-folded principals made AES hashes permanently uncrackable (`bbfff83`)

Imports stored `line.lower()`. For the AES etypes the Kerberos salt is `REALM + principal`, and the
principal's case is part of it — so a ticket for an account like `SQLSvc` was accepted, queued, sent to an
agent, and could never crack. The job ran to exhaustion and reported nothing. Silent, and live on every
deployment.

Scoped by measurement, upper-casing one field at a time in each mode's own example hash:

| Modes | principal upper-cased | realm upper-cased |
|---|---|---|
| 19600, 19700, 19800, 19900, 28800, 28900 | **0/1 recovered** | 1/1 |
| 13100, 18200 (RC4) | 1/1 | 1/1 |

RC4 keys are `MD4(password)` with no salt, which is why classic kerberoasting never hit this and the AES
modes — what modern AD hands out — do. Stored form is now what hashcat itself prints: hex lower-cased,
principal and realm verbatim. That is simultaneously crackable and round-trip stable. The unsalted RC4
modes keep the historical all-lower-case form, equally a fixed point for them, which preserves
de-duplication of the same hash pasted in different cases.

### 3. Remaining validator/hashcat parity (`0023766`)

`_KERBEROS_RE` was calibrated *from* hashcat's example hashes, so every pattern matched its own example
while rejecting shapes real tools emit and accepting shapes hashcat cannot load.

**Too strict — the hash could not be imported at all:**

| Mode | Was | hashcat accepts |
|---|---|---|
| 19800 / 19900 | exactly `{112}` | **104–112** inclusive |
| 18200 | `23$` required | etype-less `$krb5asrep$user@REALM:…` too (Rubeus, John) |

The 19800/19900 range is structural: confounder(16) + a DER-encoded `PA-ENC-TS-ENC` whose length varies
with the optional microseconds field + HMAC(12).

**Too loose — accepted at paste time, then `No hashes loaded` on the agent, with nothing to connect the
failure to what was typed:**

| Mode | Was | hashcat accepts |
|---|---|---|
| 7500 tail | any hex | exactly **104** |
| 28800 tail | any hex | exactly **32** |
| 28900 tail | any hex | exactly **64** |
| 13100 / 18200 / 19600 / 19700 `edata2` | any hex | **≥ 64** |
| 13100 triple | `\*.+\*` | nested `*` rejected |

Widening 18200 exposed a downstream break: the importer reads `line.split('$')[3]`, which becomes the
edata blob once the etype field is absent, so an etype-less AS-REP would have filed a hex string as the
username. It now takes the field carrying the `:` either way. Same block, 13100's principal arrives as
`*user`, so the leading star is stripped.

### How this was verified

Every bound came from **sweeping hashcat 6.2.6 one character at a time** across a wide range, not from
spot checks — inferring an exact length from a single example is how these patterns went wrong originally.
Worth stating plainly: **a "hashcat's example hash must validate" test would not have caught any of this.**
All nine examples passed both the old and new patterns.

- Behaviour claims tested by *cracking*, using each mode's example password, not by reading format docs.
- Every newly-accepted shape checked for round-trip stability (`--left` / `--show` re-encode through the
  same writer the crack path uses, proven by feeding them upper-case hex and watching it come back
  folded), because a stored form that differs from hashcat's output silently loses cracks.
- A corpus of ~70 vectors spanning every swept boundary plus the real tool shapes: **0 disagreements**
  with hashcat.
- An independent second sweep (1302 vectors) re-derived every bound to the same values and caught one
  false reject in the first cut of the 18200 lookahead — a machine account whose *name* is 1–3 digits
  (`18$@REALM`) is indistinguishable from an etype to a naive guard. Fixed and pinned by a test.
- Each fix mutation-tested: reverting it, inverting it, or mis-scoping it each fails a test.

### Deliberate divergences from hashcat

Documented in tests rather than left as surprises:

- **13100 bare `$krb5tgs$23$<ck>$<edata>` and John `$krb5tgs$<spn>:<ck>$<edata>`** — hashcat loads both
  and both round-trip verbatim, but neither carries a principal where the importer looks for one, so
  accepting them would file a hex blob as the username. No roasting tool in common use emits either.
- **Wrong etype in the 18200 slot** — hashcat shrugs and accepts `$krb5asrep$17$…`; we reject it. Catching
  a mode/hash mismatch at paste time is what this validator is for. (Rubeus emits etype 17/18 AS-REPs in a
  4-field shape hashcat's 18200 rejects outright, so this shape isn't producible by it anyway.)
- **Empty identity fields** (an empty 18200 principal, an empty 7500 user) — hashcat loads them; far more
  likely a truncated paste than a real hash.

### Operational note for the release

Hashes already imported in the wrong case **cannot be repaired automatically** — the original
capitalisation is gone. Re-import any AES Kerberos hashfile that has never produced a crack. This is in
the CHANGELOG entry.

### Testing

`1722 passed, 55 xfailed, 0 failed` across `tests/unit` + `tests/security`; `ruff` clean. New coverage:
the SPN shape and the SPN-less shape for both modes, the 13100-style triple still rejected, malformed
cases proving the optional group didn't become a wildcard, normalizer unit tests, an end-to-end import
asserting the stored ciphertext equals hashcat's rendering and that its md5 is the one the upload lookup
uses, exact-length probes ±2 around each pinned boundary, the full 104–112 window, the 64-hex `edata2`
floor including an odd length, the etype-optional and machine-account AS-REP shapes, and three wrong
etypes. The pre-existing vectors were truncated copies of the examples and are replaced with structurally
correct ones; hashcat's real examples are now embedded verbatim with the command to regenerate them.

### Not included

Left for separate changes rather than bundled here:

- `validate_user_hash_hashfile` only checks that a colon is present, and the importer then takes
  `line.split(':')[1]` — so a 19700 hash whose SPN contains `:1433` passes and is mis-split.
- `_HASH_ONLY_RULES` (~60 entries) and `_SHADOW_RE` were not audited for the same example-calibration bias.
  Unexamined, not cleared.
- `35300`/`35400` don't exist in hashcat 6.2.6, so their alias mapping to 13100/18200 is taken on faith
  from the existing comment, and the 18200 change propagates into 35400 unverified.


bandrel (approved):
Approving. I cross-validated this branch independently rather than reading it, and it holds up. I've also pushed some additions onto it — details and rationale below.

## How I verified it

I wrote a separate test harness against real hashcat 7.1.2 before reading this PR's implementation, then pointed it at this branch.

The oracle is `hashcat -m <mode> <hashfile> --potfile-path <empty file> --self-test-disable --left -o <out> --quiet`, which prints hashcat's own normalised form of a hash whether or not it cracks. That matters here: a mixed-case username changes the AES salt, so those vectors can never crack and an outfile-based check can't see them. Note `--potfile-disable` is rejected in combination with `--left`.

| Check against this branch | Result |
|---|---|
| `normalize_kerberos_hash` vs live echo, modes 19600 / 19700 | 12 of 12 byte-exact |
| Same, modes 19800 / 19900 / 28800 / 28900 | 16 of 16 byte-exact |
| Sweep of 240 validator-accepted SPN shapes | zero stored unnormalised |
| An independent 8-cell import-and-recovery matrix | 8 passed |
| An independent 23-test normalizer suite | 23 passed |

The scoping call in this PR is right, and I confirmed the reasoning separately: the username is part of the AES salt and the realm is not, since hashcat uppercases the realm internally. Realm case still has to be preserved, but for the other reason — hashcat echoes it verbatim, and the stored ciphertext has to match that echo byte for byte or the `md5(ciphertext)` recovery lookup misses. Worth keeping those two reasons distinct in any future doc; conflating them is an easy mistake to make.

## One bug I found, and fixed on this branch

The 19600 and 19700 patterns used `[^$]+` for the user and realm fields, so an asterisk inside either one validated. hashcat detects the SPN form with `strchr(line_buf + 13, '*')`, so any stray asterisk puts its parser into SPN mode; it then fails to find a closing delimiter and rejects the hash.

| Input shape | Validator before | hashcat 7.1.2 |
|---|---|---|
| asterisk in user | accepted | rejected |
| asterisk in realm | accepted | rejected |
| two asterisks in user | accepted | rejected |
| empty SPN `$**$` | accepted | accepted |
| valid control | accepted | accepted |

Three shapes passed at paste time and would then die on the agent with `No hashes loaded` — the same "too loose" class this PR set out to close. Fixed by tightening that class to `[^$*]+` on those two modes only. Your `{64,}` edata minimum and your `(?:\*[^*]*\*\$)?` SPN group are both correct and untouched, and the empty-SPN row above is why. 13100 is untouched, since its principal legitimately arrives as `*user`.

Each new vector is pinned properly: all six are accepted by the old `[^$]+` class and rejected by the shipped one, so they'd fail if the fix were reverted. All 47 of your existing Kerberos tests still pass.

## What I pushed onto this branch

- A self-contained live-hashcat interop CI gate: `tests/hashcat_interop/` with its own conftest override and fetch/verify script, plus `.github/workflows/kerberos-hashcat-interop.yml` running against pinned hashcat 6.2.6, 7.0.0, 7.1.0, 7.1.1 and 7.1.2. It pins seven Kerberos surfaces: both input forms accepted, SPN stripped from the echo, hex lowercased, user and realm verbatim, all-star form rejected, username salt-sensitivity, and round-trip equality between `normalize_kerberos_hash` and the live binary. That last one is the regression tripwire for this PR's whole invariant.
- The asterisk fix above, with tests.
- The two independent test files I used for validation, since they cover ground this PR didn't: a four-vector import-and-recovery matrix and a normalizer suite across every AES mode group.
- A CHANGELOG line and an allowlist entry for hashcat's published example plaintext.

The gate fails the job if pytest reports any skip, because a skipped live test exits 0 and would make the gate a no-op. It also pins `shell: bash` so `pytest | tee` doesn't swallow a failure under the default non-pipefail shell. Both of those have bitten this repo before.

Locally: `tests/unit` 1716 passed, the interop suite 26 passed with zero skips against real hashcat, ruff clean, bandit clean against the baseline, and the full pre-push suite 1863 passed.

## Why #429 stays separate

#429 covers a disjoint set of surfaces. Nothing in it overlaps the Kerberos contract, and nothing here replaces it:

- Eight offline contract tests over committed captures from five releases: status-JSON validity, the specific fields the agent's status parser reads, device info, benchmark speed parsing, outfile format round-tripping, and whether the binary still accepts every flag the server emits.
- Three live `--skip`/`--limit` slice-semantics tests, which exist as a tripwire for a change queued in hashcat's unreleased 7.1.x changelog that would alter what those flags apply to and silently shift chunk boundaries.
- Golden fixtures for 6.2.6, 7.0.0, 7.1.0, 7.1.1 and 7.1.2, plus a weekly pinned matrix and a floating newest-release check that files an issue instead of failing a PR.
- Extraction of the agent's hashcat status parsing into `agent/status.py` so it can be imported and tested at all.
- The finding that hashcat 7.0.0 emits structurally invalid `--status-json`, so agents on that release report no status whatsoever, silently.

I deliberately built the Kerberos gate here to reuse none of that, so this PR carries no dependency on #429 and can merge on its own.

The one cost of the split: once both land, two workflows each download the same five hashcat releases. Folding the Kerberos job into #429's matrix afterwards is a small, obvious follow-up, and I'd rather pay that than block either PR on the other.

bandrel:
Follow-up on my own push, for the record: I broke `e2e` and then fixed it, so if you saw it go red between `8e5874d` and `3b9233e`, that was me and not your change.

`pytest.ini` sets `testpaths = tests`, so the e2e job collects the whole tree. The e2e runner installs only `requirements-dev.txt`, since it drives the app over HTTP inside Docker and never imports `hashview.*`. My new interop module imports `normalize_kerberos_hash` for the round-trip assertion, which pulls in Flask, so collection died with `ModuleNotFoundError: No module named 'flask'` and took the job down with it.

Fixed with a directory-level `collect_ignore` guarded on `importlib.util.find_spec("flask")`, which is the same pattern #429 used for the same reason. That is not a hole in the gate: the interop workflow installs `requirements.txt`, so Flask is present and all 26 tests run there. If it ever weren't, the guard would yield zero collected tests, pytest would exit 5, and the step would fail through the pipefail shell.

All five interop legs are green against real binaries, including 6.2.6 and 7.0.0, so the Kerberos echo and salt behaviour this PR depends on holds across every pinned release, not just the 7.1.2 I had locally.

PR #480, opened 2026-09-10
Fix/issue 385 analytics account unit

i128:
## Analytics: count accounts rather than hashfile links, and scope the downloads like the page

Fixes #385. Two commits, 3 files, +364/−102 — `hashview/analytics/routes.py` gets *smaller* (166 lines
changed, mostly deletions of duplicated query logic); the rest is tests.

### The claim is right about the symptom, wrong about the cause

`corpus` was one row per `hashfile_hashes` **link row**, so a single account reached through two
hashfiles — or listed twice in one hashfile — became two people, and every plaintext-derived figure
inflated with it. The Password Reuse donut was the visible symptom: 100% reused on a scope whose
recovered passwords are all distinct, because each hash "shared" its password with its own duplicate row.

But the issue attributes that to hashes belonging to multiple hashfiles. Measured on a 1.03M-link
instance, that mechanism explains **104 of 171,149 cracked link rows**. The dominant source is **89,475
groups of identical `(hashfile_id, hash_id, username)` rows** — repeated lines in a source file, since
`hashfile_hashes` has no unique constraint on that triple.

That distinction matters twice over: it's why selecting a single hashfile was **not** a workaround
(hashfile 7 alone read 82.4%), and why a fix keyed on hashfile membership would have left it reading
82.4%.

The issue's reproduce query (`join_rows` vs `COUNT(DISTINCT hash_id)`) is also misleading as a
diagnostic. Those differ by 1.184× here, but much of the gap is legitimate — 10,192 cases are genuinely
different accounts sharing a hash, which is real reuse, not over-counting.

### Scope was never the problem; the unit was

Every `hashfile_hashes` row in range belongs in the population, and `_scoped_hash_query` already did
that correctly. What was wrong is the **unit** inside it. The corpus is now one row per account —
distinct `(Hashes.id, HashfileHashes.username)` — and `accounts` / `accounts_cracked` count the same
thing, so the page stops quoting two different "cracked" totals.

| Scope | reuse before | after | top password before → after |
|---|---|---|---|
| A customer whose file was imported twice | **100.0%** | **38.5%** | 24 → 4 |
| That customer's single hashfile | **82.4%** | **38.5%** | 12 → 4 |
| All customers | 20.6% | **20.6%** | 1285 → 1285 |
| The large single-customer scope | 20.6% | **20.6%** | 1285 → 1285 |
| `Accounts` tile, all customers | 1,028,898 | 904,862 | |

The broken scopes are fixed and **the correct ones do not move** — the figures already quoted to clients
from the large scopes are unchanged.

### Not the suggested fix — it would have been worse than the bug

The issue proposes switching the plaintext charts to a distinct-**hash** corpus. Imports dedupe hashes
globally on `(hash_type, sub_ciphertext)`, so one `hashes` row serves many accounts — on this instance
**1,048 accounts share `Training$1` through a single row**. A per-hash corpus cannot see cross-account
reuse at all and reports ~0% however widely a password is shared: it would have taken the correct 20.6%
down to **2.0%**, and the 100% case to **0%**. An over-count replaced by a near-total blind spot.

Three tests now fail if anyone implements it that way, two of which pre-date this branch.

Rows with no username collapse to one per hash rather than dropping out, so a `hash_only` import (which
carries no usernames at all) still contributes its hashes. SQL `DISTINCT` groups NULLs, which gives that
for free — `COUNT(DISTINCT a, b)` would have discarded those rows instead.

Scope of the change was tighter than the issue implies: only `corpus` and the two account tiles counted
link rows. `_recovery_by_task` and the recovery timeline already `.distinct()` on `Hashes.id`, and both
shared-password paths already built sets of usernames.

### Second commit: the downloads disagreed with the page

The page resolves scope with `_scoped_hash_query` (hashfile first, then customer). Three download routes
re-implemented that as:

```python
if customer_id:
    if hashfile_id: <this hashfile>
    else:           <this customer>
else:               <no filter at all>
```

So a hashfile id with **no** customer id fell through to the unscoped branch: the page rendered one
hashfile's charts and "Download recovered" then handed back every hash in the instance. The route
docstrings treat these args as a URL contract used by job-completion mail and the hashfile/customer
links, so such a link is producible without hand-editing a URL.

**On severity, so nobody over- or under-reacts:** `/analytics` is `@login_required` with no per-customer
authorization, so any logged-in user can already export everything by choosing the all-customers scope.
This is a correctness bug — you asked for one hashfile and got the database — not a privilege escalation.

All three now call the same helper as the page, which is the point: the defect was a second
implementation drifting from the first, so this removes the second implementation rather than patching
it.

**fig9 had a second defect.** It selected hashes with `.having(func.count() > 1)` over `hashfile_hashes`
rows, so a single account listed twice satisfied the predicate and was exported as *sharing a password
with itself* — the same link-row-vs-account confusion as the reuse donut, in a file people hand to
clients. Delegating to `_shared_groups` fixes the counting and the scoping together and deletes ~55
lines of duplicated query logic. That route isn't linked from the template, which is presumably why it
was missed when the card itself was fixed.

Behaviour otherwise unchanged: with both args set `hashfile_id` still wins (matching the page);
customer-only and unscoped requests are untouched.

### Testing

`1784 passed, 57 xfailed, 0 failed`; `ruff` clean. Six mutations tested across the two commits, all
caught: reverting the corpus distinct, switching it to `Hashes.id` (the issue's fix), reverting the
account tiles, reinstating the old download precedence on two routes, and fig9's link-row counting.

New coverage: one account in two hashfiles reports 0% reuse; triplicate rows in one hashfile count once;
genuine cross-account reuse is still reported; the password-property charts stay weighted per account (a
six-character password on three accounts weighs 3, not 1); `accounts_cracked == total_cracked`;
username-less rows count once per hash; and per-route download scoping including precedence and
cross-customer leakage.

### Deliberately not in this PR

- **The exports still emit one line per link row**, so a duplicated row appears twice in the file even
  though it now counts once in the charts. Making file contents distinct changes what users receive.
- **The `Accounts` tile still reads ~602k for a customer holding mostly `hash_only` imports**, nearly all
  of it username-less rows counted one per hash. Honest, arguably mislabelled — but the alternative shows
  ~56 accounts for a customer holding 600k hashes, which seems worse. A labelling decision, not a
  counting one.
- **A unique constraint on `hashfile_hashes (hashfile_id, hash_id, username)`** would stop the 89,475
  duplicate-row groups at the source instead of compensating in the read path. That's the deeper fix and
  wants its own evaluation, including what to do with the duplicates already stored.


i128:
LMK if this addresses your concern @bandrel

Issue #383, opened 2026-08-14
Rule row survives when its file is gone: /v1/rules/<id> 404s and agents retry the download forever


bandrel:
## Summary

When a rule's file disappears from `control/rules/` but its `Rules` row survives, nothing on the server notices. The row stays in the catalog and in `GET /v1/rules`, `GET /v1/rules/<id>` answers `404 {"msg": "Rule file missing on disk"}`, and every agent retries that download on **every** sync, forever, without ever converging or surfacing the problem to an admin.

## Reproduction

Found while testing on the docker dev stack, where `control/` isn't on a volume — so recreating the `app` container wipes the uploaded files while the DB volume keeps the rows:

```
$ curl -b "uuid=$KEY" -X POST --data-binary @corp_8.hcmask \
    http://localhost:5000/v1/rules/add/corp_8.hcmask
{"msg":"Rule added","rule_id":2,"status":200,"type":"message"}

$ docker compose up -d --force-recreate app      # control/rules/ is gone, rules row isn't

# agent, next sync:
[INFO]    Downloading new rule 2.
[WARNING] GET http://app:5000/v1/rules/2 returned HTTP 404: {"msg":"Rule file missing on disk", ...}
[WARNING] No data received for rule 2; will retry next sync.
```

The container-recreate part is a dev-stack artifact, but any path that loses the file keeps the row — a botched restore, a manual `rm`, a half-failed upload, a volume remount.

## Why it doesn't self-correct

- `api/routes.py` `v1_api_get_rules_download` returns 404 when `src_path` is missing, but nothing marks the `Rules` row.
- `install/hashview-agent/hashview-agent.py` `_install_rule_file()` returns `None` on a failed download, so the rule is never written into `new_manifest` — and `sync_rules()` re-attempts it on the next call, indefinitely. That's the right conservative behavior for a transient failure; it just never escalates a permanent one.
- The rule stays listed in `GET /v1/rules` and in the web UI's rule pickers, so it can still be attached to a new task.

## Impact

The worst case isn't the log noise — it's that a task can reference a rule whose file doesn't exist. `build_hashcat_command` emits `-r control/rules/<name>` regardless, so the agent runs hashcat against a path it could never fetch and the job fails at crack time, with the root cause several layers away from the error.

## Suggested handling

Any one of these would close the loop; roughly in order of effort:

[...]

Issue #398, opened 2026-08-18
API: no DELETE endpoint for /v1/wordlists/<id> (wordlists can be created but never deleted via API)


bandrel:
## Summary

The v1 API can **create** wordlists (`POST /v1/wordlists/add/<wordlist_name>`, `hashview/api/routes.py:863`) but provides no way to **delete** one. The only other wordlist routes are `GET /v1/wordlists` (`routes.py:783`), `GET /v1/wordlists/<id>` (`routes.py:797`) and `GET /v1/updateWordlist/<id>` (`routes.py:835`) — there is no `DELETE /v1/wordlists/<int:wordlist_id>`.

Jobs and hashfiles are both fully deletable via the API (`DELETE /v1/jobs/<int:job_id>`, `routes.py:960`; `DELETE /v1/hashfiles/<int:hashfile_id>`, `routes.py:1467` — added for #310). Wordlists have the equivalent web-UI route (`POST /wordlists/delete/<int:wordlist_id>`, `hashview/wordlists/routes.py:179`) but nothing on the API surface.

## Impact

`POST /v1/wordlists/add` does not check the name against the catalog — it appends a `Wordlists` row on every call. So an API client that uploads a wordlist twice creates two rows sharing a name, and **neither can be removed without web-UI access**.

This is not hypothetical: a duplicate-wordlist mishap on a live instance previously had to be cleaned up by hand through the UI, precisely because the API offers no delete. Any client doing name-based resolution then has to deal with an ambiguous name it cannot fix.

Wordlists make this more costly than rules, because they are large. An accidental or unwanted upload consumes real disk on the server (stored gzip -9 at `control/wordlists/`) with no programmatic way to reclaim it, which matters for automated tests and for any tooling that copies catalogs between instances.

## Suggested fix

Add `DELETE /v1/wordlists/<int:wordlist_id>` mirroring `v1_api_delete_job` (`routes.py:960`) and the web-UI `wordlists_delete` guards (`wordlists/routes.py:179`):

- Require an authorized user (`is_authorized(user=True, agent=False, ...)`) and resolve the caller via the `uuid` api_key cookie.
- Return `404` when the wordlist does not exist.
- Enforce ownership: `user.admin or wordlist.owner_id == user.id`, else `403`.
- Refuse deletion of dynamic wordlists, as the UI does (`wordlists/routes.py:191-195`) — they are regenerated from the DB and the row is not the user's to remove.
[...]

PR #492, opened 2026-09-11
Fix/catalog missing file 383

i128:
Closes #383.

## The problem

A `Rules` or `Wordlists` row can outlive its file on disk, and nothing on the server noticed. The row stayed in every listing and every task picker, `GET /v1/rules/<id>` answered a JSON 404 that only the agent's own log ever saw, agents re-requested the dead file on every task assignment forever with no backoff, and a task could still be built against a file that can never be fetched.

When hashcat finally failed on it, `run_hashcat` answered the non-empty stderr with `os.kill(os.getpid(), SIGINT)` — killing the agent and leaving the JobTask wedged in `Running` — and the only admin-visible artifact was an email containing raw hashcat stderr.

## What changed

**Detection.** `utils.resolve_control_file(stored_path, subdir)` plus `rule_file_missing` / `wordlist_file_missing` and the bulk `missing_rule_ids` / `missing_wordlist_ids`. Six sites each decided this independently and **disagreed**: `/v1/rules/<id>`, `/v1/wordlists/<id>` and `remove_rule_file` resolved by basename against `control/<subdir>`, while `rules_download`, `wordlists_download` and the tasks-list byte-size lookup stat'ed the raw stored path — which only works when the process CWD happens to be the repo root. Seeded rows (`hashview/control/rules/best64.rule`) store a relative path, so those three were latently broken under any non-repo-root `WorkingDirectory`. All six now share the basename-confined form, which is also what keeps a crafted or legacy path from reaching outside the control directory.

**Dynamic wordlists are never reported missing.** Their file is a regenerable cache — `update_dynamic_wordlist` rebuilds it per download and never reads `wordlist.path` — and every seeded `dynamic-*.txt` is a zero-byte placeholder, so both absence *and* zero length are the expected state. A naive `getsize() > 0` probe would false-alarm on ten rows on every install; there is a test pinning that.

**API.** `GET /v1/rules` and `GET /v1/wordlists` carry a `missing` boolean per row, grafted after serialization (setting it on the ORM instance *would* serialize — `AlchemyEncoder` walks `dir(obj)` — but it pollutes the identity map). Always emitted, so an absent key unambiguously means "server predates the flag". `tests/unit/test_issue_xfail_stale_catalog_file.py` was the `xfail` spec for exactly this; both tests XPASS, so the markers are dropped and the file becomes a live regression test.

**Admin alerts.** An hourly `CATALOG_HEALTH` job reports missing catalog files and reports them again on recovery. The alert is **aggregated** — one message per sweep per direction, not one per row: `notify_admins` reaches email, a Pushover *phone push* and Slack, and these failures are almost always correlated (a volume move, a botched restore). The body names each row and the tasks referencing it (`wl_id_2` included), which is the field that decides re-upload vs delete. The audit log still gets one event per row, since the Logs viewer parses `target` into entity/id/name. De-duplication uses a new `file_missing_notified` column mirroring `agents.offline_notified` — it records whether the admins have been *told*, never whether the file is there.

A circuit breaker skips the sweep entirely when a control directory is not a directory, so an unmounted volume cannot flag the whole catalog, latch every row, and fire an equally huge "restored" alert on remount.

**UI.** Red `FILE MISSING` badges on both listings and on task rows (answers "which of my tasks are broken?" without opening anything), disabled downloads, a real size-on-disk in the rules info modal (it was a hardcoded em dash), and delete dialogs that list the blocking tasks *and their jobs* with links — turning "can not be deleted" from a dead end into a procedure.

**Pickers.** Missing entries are excluded when adding a task — at the **route**, so a hand-crafted POST fails choice validation too — and refused by `POST /v1/tasks/add`, which has no picker to grey out. Editing is the opposite: the task's current value is kept and labelled, because excluding it would leave the select on `selectedIndex -1` and the task could never be saved again. That required widening the current-value prepend, which covered only attackmode 0 and **never covered `wl_id_2` at all**, so combinator and hybrid tasks would otherwise have become uneditable.

**Restore in place** — the answer to the hard case. When tasks already reference a stranded row, deletion is (correctly) refused, and re-uploading did not help: both `rules_add` and `ingest_static_wordlist_file` always mint a new row with a fresh `token_hex(8)` path, so a re-upload orphaned the reference *further*. The startup backfill has been logging *"re-upload the wordlist to restore it"* for a condition that had no such remedy. `POST /rules/<id>/restore` and `POST /wordlists/<id>/restore` now write into the existing row, keeping its id and path basename — so every `Tasks.rule_id`/`wl_id`/`wl_id_2`, every already-materialized `JobTasks.command` (the agent-side path is the basename, so it is unchanged) and every `Hashes.task_id` attribution stays valid. Both stage under `control/tmp` and `os.replace()`, so a corrupt upload cannot destroy a file that is still good. `rules_view` becomes the small-file variant: its unconditional read ran *before* the POST branch, so a missing file bailed and the write was unreachable.

Wordlist restore is offered on **every** static row, not just stranded ones — it is also the only way to update a static wordlist's contents, since there is no wordlist edit route at all. The dialog states outright that importing from `control/wordlists_import/` will *not* restore the entry, because that path calls the same ingest and would silently leave the row broken.

## Bugs found along the way

- **`wordlists_delete` never checked `Tasks.wl_id_2`**, so a combinator task's second wordlist could be deleted out from under it — even though the wordlists listing and `build_hashcat_command` both count it. Now refused for either reference. Deletion is strictly stricter; the flash string is byte-identical so existing assertions hold. *Note for #398: the future `DELETE /v1/wordlists/<id>` must mirror **this** guard, not the old one.*
- **`sync_rules` destroyed the agent's good local copy on any failed download.** The id dropped out of the manifest and `_prune_orphan_files` deleted the file; the next task using that rule made hashcat fail, which SIGINTs the agent and wedges the JobTask. A single transient 502 could take a rig down. `sync_wordlists` has always guarded against exactly this. The fix is ordered: the superseded file is removed only *after* a successful install and only when the filename changed.
- **`/rules/delete/<id>` accepted GET.** With no global `CSRFProtect`, a single `` on any page an admin visited was enough to delete a rule. Every legitimate caller already POSTs. (`jobs_delete` has the same shape and is deliberately left for its own pass.)

## Deployment notes

- **Migration `e5d1c7b3a904`** on head `f3b8c1a7d942` — additive, guarded, adds `file_missing_notified` to `rules` and `wordlists`. Existing rows land on `0`, so the first sweep after upgrade sends one alert naming everything already broken. That is intended: it is the "tell me what is already wrong" report. On the reference instance that is a single line for wordlist 6.
- **Agents should be redeployed.** Not for the `missing` flag — that is wire-compatible in both directions (an old agent ignores the extra key; a new agent against an old server sees no key and behaves as today), which is why there is deliberately **no version bump**: bumping `__version__` without redeploying every agent would 426 the whole fleet. The redeploy is for the prune fix, which is where the rig-death failure mode lives.
- Deploy server-first, per house convention.

## Verification

`2173 passed, 72 xfailed, 0 failures` across `tests/unit`, `tests/agent_unit` and `tests/security`; ruff clean.

New: `test_catalog_file_missing.py` (resolver, including the relative-seeded-path case under `monkeypatch.chdir` — precisely what the old raw `os.path.exists` got wrong — and the dynamic false-alarm cases, absent *and* zero-byte), `test_catalog_health_check.py` (latch, recovery, aggregation, task ids, audit events, circuit breaker, job registration), `test_catalog_missing_file_ui.py` (badges, pickers, the hand-crafted POST, the combinator `wl_id_2` re-save), `test_catalog_restore.py` (id/path preserved, references intact, write confinement, invalid gzip leaves the old file), `tests/agent_unit/test_sync_missing_flag.py` (7 of its cases fail on `v0.8.3-dev`).

A read-only probe of the production database and filesystem — no `create_app`, no ORM — flags exactly wordlist 6 and nothing else: no rules, no dynamic false alarms, Best64's relative path resolving correctly, 0 referencing tasks.

The first commit is a pure test refactor with no behaviour change: ~2 dozen unit files build catalog rows with paths that never existed, and five of them POST to `/tasks/add` or `/tasks/edit`, so they had to become file-backed before the pickers could filter. It is droppable-first and reviewable on its own.

## Deliberately out of scope

Each of these deserves its own issue rather than riding along here:

- **`build_hashcat_command` appends an empty-string argv element** when a row is gone entirely (`utils.py:1601-1616`) — the exact argv that leads to the agent SIGINT above.
- **`run_hashcat`'s `os.kill(os.getpid(), SIGINT)`** on any hashcat stderr. This PR removes the most common *trigger*, not the mechanism.
- **`DELETE /v1/wordlists/<id>` (#398)** — it has its own xfail spec pinning a different envelope and auth matrix.
- **Blocking job start / task dispatch** on a missing file. #383 asks to surface, not enforce; gating a queued job from starting is a policy change with real blast radius.
- **Checksum drift** (file present, sha256 no longer matches) — a strict superset, and rehashing a 53 MB `.gz` per sweep is a different cost class.
- **Auto-repair or auto-delete of stale rows** — an admin who mistakenly unmounted a volume would lose their catalog.


bandrel:
## Review: blocking issues plus four smaller ones

Reviewed at `0989e9e`. Every finding below was reproduced or read directly out of the branch, not inferred from the description. The description itself is excellent and made this much faster to review.

---

### 1. Blocking: `migration-e2e` is red

Both hardcoded head pins still point at the old revision, so the job times out:

```
ERROR: timed out waiting for alembic at dev head (f3b8c1a7d942) (last value: 'e5d1c7b3a904')
```

```diff
--- a/tests/integration/test_migration_e2e.py
+++ b/tests/integration/test_migration_e2e.py
-DEV_HEAD = "f3b8c1a7d942"
+DEV_HEAD = "e5d1c7b3a904"

--- a/tests/run_migration_e2e.sh
+++ b/tests/run_migration_e2e.sh
-DEV_HEAD="f3b8c1a7d942"
+DEV_HEAD="e5d1c7b3a904"
```

Worth considering the dynamic-head approach this PR already applied to `test_migration_drift_idempotency.py`, since these two pins have to be touched again the moment #493 merges.

---

### 2. Blocking: most task edits now fail silently

The edit modal and the edit route disagree about which wordlists exist, and the hidden second-wordlist select is caught in the middle.

* The modal passes `keep_missing=True`, so `templates/tasks.html.j2:73` renders **every** missing wordlist into `etk-wl2`.
* `tasks/routes.py:352` re-admits the task's own `wl_id_2` **only for attackmode 1**, and line 372 drops every other missing row from the choices.
* `templates/tasks.html.j2:376` only calls `setSel('etk-wl2', ...)` when the task actually has a second wordlist. For any other mode the hidden select keeps, and submits, its first rendered option.

When the lowest-id wordlist is the stranded one, that first option is a value the form rejects. Reproduced on a clean checkout of this branch, editing a plain dictionary task:

```
FIRST etk-wl2 OPTION -> 1 (gone=1 healthy=2)
STATUS -> 200
Not a valid choice present -> True
name after POST -> plain          <- the rename was discarded
```

This is a genuine regression. On `v0.8.3-dev` both selects are populated unconditionally, so the template and the route agreed. It is also the same "task becomes un-saveable" failure this PR set out to prevent, arriving through a field the current-value prepend does not cover.

`tests/unit/test_catalog_missing_file_ui.py:196` encodes the wrong assumption in a comment, saying the hidden select submits the first *present* wordlist. It only passes because the healthy wordlist happens to get the lower id.

The template cannot filter per-task, since the edit modal is filled client-side and does not know which task will open. So the fix belongs on the server: validate only the selects the submitted attack mode actually uses.

```diff
--- a/hashview/tasks/routes.py
+++ b/hashview/tasks/routes.py
@@ def task_edit(...)
+        # The modal's hidden selects still submit, and the edit modal deliberately
+        # renders missing rows (keep_missing=True) that the choices above exclude.
+        # Only the fields the submitted attack mode actually uses are meaningful,
+        # so don't choice-validate the rest -- otherwise a mode-0 edit is rejected
+        # by whatever etk-wl2 happened to show.
+        try:
+            submitted_mode = int(request.form.get('hc_attackmode', task.hc_attackmode))
+        except (TypeError, ValueError):
+            submitted_mode = task.hc_attackmode
+        if submitted_mode not in (0, 1, 6, 7):
+            tasksForm.wl_id.validate_choice = False
+        if submitted_mode != 1:
+            tasksForm.wl_id_2.validate_choice = False
+        if submitted_mode != 0:
+            tasksForm.rule_id.validate_choice = False
+
         if tasksForm.validate_on_submit():
```

The per-mode branches in the POST body already only persist the fields that mode owns, so nothing unvalidated reaches the database.

Please also add a regression test that puts the missing wordlist at the **lower** id and asserts a mode-0 and a mode-3 edit still redirect. That ordering is the whole bug.

---

### 3. A notification transport failure re-sends the same alert every hour, forever

In `_catalog_health_check_inner`, `notify_admins` runs at line 562, the `file_missing_notified` latch is set at lines 569 and 574, and there is a single `db.session.commit()` at line 600.

Neither `notify_admins` nor `send_pushover` catches anything, and `send_pushover` does a bare `requests.post(...)` followed by `response.json()`. A Pushover or Slack timeout, a DNS failure, or a 5xx with a non-JSON body raises *after* the emails have already gone out. Nothing latches, nothing commits, and the identical message fires again next sweep.

The existing `agent_health_check` has the same ordering but commits per row, so it loses at most one latch. This sweep loses the whole batch, which is a direct consequence of the aggregation design.

```diff
-        notify_admins(
-            'Hashview: %d catalog file%s missing on disk' % (total, '' if total == 1 else 's'),
-            '\n'.join(body))
+        try:
+            notify_admins(
+                'Hashview: %d catalog file%s missing on disk' % (total, '' if total == 1 else 's'),
+                '\n'.join(body))
+        except Exception:
+            # A dead Pushover/Slack endpoint must not un-latch rows whose email
+            # already went out, or this aggregated alert repeats hourly forever.
+            logger.exception('CatalogHealthCheck: admin notification partially failed; latching anyway.')
```

Same treatment for the restored branch. Making `send_pushover` and `_post_slack` never-raise, the way `send_email` already does, would be the broader fix and would help `agent_health_check` too.

---

### 4. The recovery alert tells admins every restored row is unused

```python
body += _catalog_alert_lines(restored_rules, {}, 'rule')
body += _catalog_alert_lines(restored_wordlists, {}, 'wordlist')
```

`_catalog_alert_lines` prints `not referenced by any task` whenever the map has no entry, so every restored row gets that line regardless of truth. The description calls the task list the field that decides re-upload versus delete, so this is the one line that matters being wrong every time. Either pass the real map or suppress the task line for the restored direction.

---

### 5. `resolve_control_file` should test `isfile`, not `exists`

Confinement genuinely holds. I checked that `os.path.basename('../../etc/passwd')` is `passwd`, so traversal is neutralised as claimed. The gap is directory-shaped basenames:

| stored path | basename | resolves to | `exists` |
|---|---|---|---|
| `../../etc/passwd` | `passwd` | `control/rules/passwd` | correct |
| `/x/..` | `..` | `control/` | **true** |
| `a/b/` | `` | `control/rules/` | **true** |

Those rows report healthy and then fail later in `getsize` or `os.replace`.

```diff
-    return target if os.path.exists(target) else None
+    return target if os.path.isfile(target) else None
```

---

### 6. The write path this PR taught to create files still has no CSRF protection

The description says restore uses small FlaskForms so `validate_on_submit` gives real CSRF validation. That is true of both new restore routes, which I verified. It is not true of `rules_view`, which this same PR upgraded from "edit an existing file" to "create one":

```python
new_content = request.form.get('content')
```

With no global `CSRFProtect`, that is a cross-site-writable owner-or-admin file write, and its blast radius grew in this PR. Worth giving it the same small FlaskForm treatment as the restore routes rather than leaving it for a later pass. Separately, `os.path.basename(rule.path or '')` has no empty-basename fallback here, unlike `rules_restore` which mints a random name.

---

### What I checked and found clean

* **Agent sync reordering.** Traced `_install_rule_file`, the manifest update and `_prune_orphan_files`. The new order leaks nothing and leaves no manifest entry pointing at an absent file. The `elif local_entry` branch correctly re-records the old entry, whose filename is what is actually on disk.
* **Wire compatibility.** `entry.get('missing')` is falsy against an old server and the checksum is unchanged, so an old agent with a local copy does not even attempt a download. The no-version-bump call is right.
* **XSS.** Autoescape is on for `.j2`, the diff adds no `|safe`, no script interpolation and no `innerHTML`, and the new JS uses `textContent`.
* **Test helper teardown.** `cleanup_control_files` unlinks only the exact randomly-named paths it created. No glob, no directory sweep, so there is no risk to a developer's real catalog.
* **OpenAPI.** Matches what the code emits, including the always-false dynamic case.
* **Migration.** Guarded, correct parent, works on both backends. See the cross-PR note below.
* **`assert present.id`** at `test_catalog_missing_file_ui.py:147` is a no-op and asserts nothing.

---

### Cross-PR: migration head collision with #493

Both `e5d1c7b3a904` here and `a4c9e7b21f60` there declare `down_revision = 'f3b8c1a7d942'`. Whichever merges second produces two alembic heads and needs its parent re-pointed and the head pins bumped again. #493 flags this in its description; worth mirroring here since this is also the PR that missed the pins.

bandrel:
Pushed fixes for everything in the review above, one commit per issue.

| commit | issue |
|---|---|
| `8a07b66` | migration-e2e head pins, which is what made CI red |
| `d434704` | the task-edit regression |
| `e95b7fd` | alert latch unwinding on a transport failure |
| `53d4539` | restored alert claiming every row is unused |
| `95efa46` | `isfile` instead of `exists` in the resolver |
| `f57beb7` | CSRF on the rule-file editor |
| `9458222` | the no-op `assert present.id` |

Each behavioural fix ships a test that **fails without it**. I checked that by reverting the change under the test rather than trusting that a green test proves anything:

* the two task-edit tests fail with the validation change removed
* both latch tests fail with the `try/except` removed
* the resolver test fails against `os.path.exists`
* the CSRF test rewrites the file when the route reads `request.form` again

On the task-edit fix specifically: the regression test creates the stranded wordlist **first**, so it takes the lower id and lands as the first option in the hidden select. That ordering is the entire bug, and it is why the existing tests passed. There is a second case for mask mode, where both wordlist selects submit excluded values.

The CSRF commit adds `RuleContentForm` rather than reusing the already-defined-but-unused `RulesEditForm`, which carries a `DataRequired` name field the editor template never posts. I left that dead class alone rather than widening the diff. No `DataRequired` on the content field either, so what the route accepts is unchanged; only who can drive it.

Full offline suite: **2184 passed, 72 xfailed, 0 failures**, ruff clean. The two skips are `mysqldump not installed` on my machine, not CI.

Two notes:

* `tests/security/test_csrf_xfail.py` still xfails, so nothing there XPASSes from the editor fix. The single remaining xfail is untouched.
* The migration head collision with #493 is unchanged. Whichever merges second still needs its `down_revision` re-pointed and both pins bumped again. #493 has now been pushed to as well, so its pin is at `a4c9e7b21f60`.

PR #493, opened 2026-09-11
Fix/mask option sentinel

i128:
﻿## The problem

A mask whose first character is `-` reaches hashcat as an argv element that its **option** parser claims before the mask parser ever sees it:

```
$ hashcat -m 0 -a 3 <hashfile> '-?l?l'
hashcat: invalid option -- '?'
Invalid argument specified.                                     rc=255
```

Distributed chunking manufactures these automatically. `chunking._expand_mask` splits a mask task by expanding the leading mask position over its character set, and `?s` — so also `?a` — contains `-`.

**It is not actually a chunking bug.** A mask an operator simply types as `-?d?d?d` fails identically today with no chunking involved. The chunker just guarantees it happens.

## What is actually broken — measured, not assumed

All against hashcat 6.2.6, as real cracks (`--stdout` has a different positional signature and gives misleading results).

**Only `-` breaks.** Sweeping all 95 characters of `?a` as a chunk prefix, exactly one fails. Two things in the original report turned out not to be true:

- `??a?a?a?a?a?a?a` is **not what the code emits**. `_escape_mask_literal` already escapes `?` → `??`, so that chunk is `???a?a?a?a?a?a?a`, and hashcat parses it correctly (`???a?a` generates 9025 candidates, as intended).
- Conditional quote-wrapping **cannot work at all**. `build_hashcat_command` returns an argv *list* and the agent runs `Popen(argv, shell=False)`. With no shell in the path, a `'` becomes a literal mask character rather than a quote.

**But the blast radius is larger than one chunk in 95.** When a plan expands more than one mask position, every `-X` combination is generated. Running the argv the server actually builds, over the planner's actual output:

```
?s?d  ->  330 chunks
  with the sentinel : 330/330 run clean
  without it        : 320/330
      -0  rc=255  hashcat: invalid option -- '0'
      -1  rc=255  hashcat: option requires an argument -- '1'
      -2  rc=255  hashcat: option requires an argument -- '2'
      -3  rc=255  hashcat: option requires an argument -- '3'
      -4  rc=255  hashcat: option requires an argument -- '4'
      -5..-9      hashcat: invalid option
```

`-1` through `-4` are the worst of them: those are hashcat's own custom-charset flags, so they do not merely fail — they would consume whatever argv element followed.

## Why not `--skip`/`--limit`

The Hashtopolis model was considered and rejected. Hashtopolis never touches the mask; it appends `-s`/`-l`. It can only do that because an agent runs `hashcat --keyspace` and posts the integer back, and that round trip is unavoidable:

| mask | `hashcat --keyspace` | true candidates |
|---|---|---|
| `?a`×8 | **7,737,809,375** (= 95⁵) | 6,634,204,312,890,625 |
| `?l`×5 | 456,976 | 11,881,376 |
| `?l`×6 | **456,976** | 308,915,776 |
| `?l`×7 | **456,976** | 8,031,810,176 |

Three different attacks, one keyspace — hashcat decides how much of the mask goes to the device-side loop, so the server can never derive the unit. Adopting it would mean a new heartbeat state, a new endpoint, chunk planning that can no longer happen at queue time, and restoring the `job_tasks.Keyspace`/`keyspace_pos` columns that were deleted in 2021 after repeated keyspace-calculation bugs. `chunking.py`'s module docstring makes "no hashcat is invoked on the server" an explicit axiom. Hashtopolis still carries open issues for exactly this machinery (measure failures, negative keyspace, integer overflow at 10 mask positions).

## The fix

Emit hashcat's `--` end-of-options marker before the positionals, so a mask that starts with `-` is never read as an option. Verified across every mask mode:

```
-a 3 -- <hashfile> '-?l?l'               runs        rc=1
-a 3 -- <hashfile> 'a?l?l'               cracked     rc=0
-a 6 -- <hashfile> <wordlist> '-?d'                  rc=0
-a 7 -- <hashfile> '-?d' <wordlist>                  rc=0
```

Two supporting pieces:

- **`split_mask_field()`** returns the same token list `mask_argv()` always did, *plus* which element is the positional mask — which is what lets the caller hoist options ahead of the sentinel. `--` swallows everything after it, so a custom charset left behind it yields `Custom-charset 1 is undefined.` It deliberately does **not** scan for options from index 1 unconditionally: that would collapse the tested `-1 ?u?l?d ?1?1?1?1?1` form back into one element, reintroducing the bug #491 just fixed. It also slices the mask out of the original string by character offset rather than rejoining tokens, so a mask beginning with a literal space survives.
- **`mask_attack_argv()`** assembles the mode 3/6/7 tail, so the live test can build the real token order instead of re-implementing and drifting from it.

### Emission is conditional, and that is the whole compatibility story

Everything after `--` is a positional, and un-upgraded agents append `--status-json` to whatever the server sent. Measured:

| | |
|---|---|
| mode 3, `--status-json` trailing after `--` | works — the extra positional is ignored |
| mode 6, trailing | **rc=255 `?a?a?a?a: No such file or directory`** |
| mode 7, trailing | **rc=255 `--status-json: No such file or directory`** |
| `-a 6 --stdout -- <wl> '?d' --status-json` | 4 candidates instead of 20 — silent corruption |
| `--status-json` placed after `--` | **not honoured**: 0 status objects vs 23 before it |

So emitting `--` unconditionally would have broken **every mode 6 and 7 command in existence — chunked or not — on any agent that has not been upgraded**, because the server cannot know which agent will run it. That is a two-attack-mode outage until the last agent is updated.

Emitting it only when the mask would actually be misread leaves every other command byte-for-byte identical — proven by the entire existing suite passing unedited — so the only behaviour delta in the system is on chunks that are already 100% broken.

`--status-json` is emitted paired with the sentinel, so a mode-3 chunk keeps reporting status on an old agent (whose trailing duplicate then lands after `--` as an ignored positional). The agent change inserts its own copy ahead of the sentinel and skips it when the server already sent one.

**Compatibility matrix** — only `-`-mask commands differ; every other cell is byte-identical:

| | old agent | new agent |
|---|---|---|
| **old server** | today: rc=255 invalid option | identical to today |
| **new server** | mode 3: **runs, cracks, telemetry kept**. modes 6/7: still broken, different message | modes 3/6/7 ✓, telemetry ✓ |

No cell regresses.

## Adjacent bugs fixed

- **The e2e crack shim mis-parsed `--skip`/`--limit` and `-1`..`-4`.** It dropped the flag via the `startswith("-")` catch-all but left its *value* in the positional list, so a chunked command parsed as hashfile `"0"`. This has made end-to-end coverage of any chunked run unsound — for **wordlist** chunks as much as mask ones — independently of this change. It now also models `--`.
- **`JobTasks.chunk_mask` widened 64 → 255.** Defensive, not a live bug: a sub-mask is never longer than the mask it came from and `Tasks.hc_mask` is `VARCHAR(50)`, so 64 suffices today. The hazard is an undocumented coupling between two column widths in different tables with 14 characters of headroom — widen `hc_mask` (the form has no length validator at all) and the chunker starts writing *truncated* sub-masks, which are still **valid** masks, so nothing errors and the wrong keyspace is cracked silently. A test now pins the never-grows invariant.

## Verification

`2122 passed, 74 xfailed, 0 failures` across `tests/unit`, `tests/agent_unit` and `tests/security`; ruff clean; migration round-tripped upgrade → downgrade → upgrade.

**No existing test needed editing.** That was the acceptance criterion for conditional emission, and any test that had broken would have meant the byte-identity property was violated.

The new `tests/hashcat_matrix/test_mask_chunk_coverage.py` is the gap that let this ship: `test_chunk_coverage.py` covers only the wordlist `--skip`/`--limit` path, and the unit tests assert a generated sub-mask is valid using **Hashview's own** parser — `parse_mask('-?d?d')` is perfectly happy with it. So every assertion there runs real hashcat over the real planner output through the real argv assembler; nothing is hand-written, because a hand-written list would drift from what production emits, which is precisely the failure mode being fixed. It covers: every sub-mask of a full plan runs; a negative control proving the sentinel is what makes that true; exactly-once target coverage (no gap, no double coverage); the sentinel in each of modes 3/6/7, which put the mask in different positional slots; and that hashcat honours `--status-json` before the sentinel but silently ignores it after — including a live check of the back-compat claim that an old agent's trailing duplicate is harmless in mode 3.

It uses `?s` (33 chunks) rather than `?a` (95) to stay inside the 20-minute CI budget across five pinned releases — identical code path, identical character — and carries no `skipif` or `xfail`, since the workflow greps for `skipped` and fails the job. 3m41s locally against 6.2.6.

## Deployment notes

- **Agents should be redeployed.** Modes 6 and 7 with a `-`-leading mask need the updated agent; mode 3 works on an existing one. **No version bump**, deliberately: `versionCheck` 426s any agent older than `__version__`, so bumping would force a fleet-wide outage in order to fix commands that are already broken on every version. Let it ride the next release.
- **This branch adds migration `a4c9e7b21f60` on head `f3b8c1a7d942`** — the same head `fix/catalog-missing-file-383` chains off. Whichever merges second will have two heads and needs its `down_revision` re-pointed. The widening is the last commit here and is severable if that is inconvenient.

## Deliberately out of scope

- **The `--keyspace` + `--skip`/`--limit` model.** Settled above. Worth recording for whoever revisits it: mask chunking is also very **coarse** today — chunk counts quantize to charset-size products, so `?a`×8 yields exactly 95 chunks of 95⁷ candidates each (~19 h apiece at 1 GH/s against a 1 h target). Granularity, not this bug, is the real argument for skip/limit.
- **Making modes 6/7 `-` masks work on un-upgraded agents** — impossible with `--`. The lever would be having `_expand_mask` expand a *non-leading* position when the leading charset contains `-` (coverage is mathematically identical), which needs no argv or agent change at all. Worth an issue if that ever becomes a hard requirement.
- **`--` for modes 0 and 1** — no mask, and every positional is a server-generated `control/...` path.
- **Giving custom charsets their own form fields**, which is the actual root cause of the whole `mask_argv` heuristic; its own docstring already says so.
- **Widening `Tasks.hc_mask`** or adding a length validator to the task form (there is none today).


bandrel:
## Review: core fix is sound, two things to change before merge

Reviewed at `75b69df`. CI is green. The `--` approach is right, the reasoning for making emission conditional is correct and well evidenced, and the measurement-first framing made this genuinely pleasant to review. The live matrix test is real coverage with negative controls that actually bite.

Two substantive issues and a few small ones.

---

### 1. A chunked custom-charset mask ships as one literal positional

`split_mask_field` disables its leading-option guard when `first_token_is_mask=True`, which is exactly the path every chunk takes:

```python
if (not first_token_is_mask
        and tokens[0][1].startswith('-')
        and _LEADING_OPTION.match(tokens[0][1])):
```

For a task mask written options-first, which is the form #491 added support for, the planner still chunks it and the whole field collapses into a single mask token. Verified against the real planner and the real assembler on this branch:

```
mask          '-1 ?u?l?d ?d?d?d'
chunks         26
first sub-mask '-1 A?l?d ?d?d?d'
argv           ['-a', '3', '--status-json', '--', 'h.txt', '-1 A?l?d ?d?d?d']
```

hashcat accepts that and brute-forces a fifteen-character literal mask containing `-1` and two spaces. No error, wrong keyspace, chunk coverage silently lost. The `-1` charset flag also ends up *after* the sentinel, which the docstring correctly says must never happen.

Two notes on scope, in fairness to the change:

* This is **not a clean regression**. `chunking.py` is byte-identical to base, and the root cause is pre-existing: `_expand_mask` treats the charset *definition* as the leading mask position and expands that. On base the same chunk produced `['-1', 'A?l?d', '?d?d?d']`, which is also the wrong keyspace, just a differently wrong one.
* But the docstring claims the options-first form behaves "exactly as before", and that is false for chunks. And the new form is worse in one respect: hashcat now silently accepts it as a literal mask instead of at least applying a charset.

The clean fix is upstream of the argv layer. A field carrying options is not a bare mask and should never have been chunked at all. That makes `from_chunk=True` safe by construction, because it then only ever sees a single bare token, and it fixes the pre-existing wrong-keyspace bug at the same time.

```diff
--- a/hashview/utils/chunking.py
+++ b/hashview/utils/chunking.py
@@ def _expand_mask(mask, desired_chunks, max_chunks):
     """Split ``mask`` into sub-masks by expanding leading charset position(s).

     Returns a list of >= 2 sub-mask strings, or ``None`` to signal "don't split".
     """
+    # A field carrying hashcat options ('-1 ?u?l?d ?d?d?d', or a mask with a
+    # trailing charset) is not a bare mask: parse_mask reads the charset
+    # DEFINITION as mask text and we would expand that, yielding a sub-mask with
+    # a mangled charset and the wrong keyspace. Refuse to split; run it whole.
+    # Deferred import: hashview.utils.utils imports plan_chunks from this module.
+    from hashview.utils.utils import split_mask_field
+    parts, mask_index = split_mask_field(mask)
+    if mask_index != 0 or len(parts) != 1:
+        return None
+
     tokens = parse_mask(mask)
```

I'd avoid simply applying `_LEADING_OPTION` on the chunk path. It would misfire on a legitimate sub-mask: a task mask of `?sa?d` expands to a chunk of `-a?d`, and `-a` matches that pattern.

Worth a test asserting that a mask containing a custom charset plans exactly one chunk.

---

### 2. Whitespace regression: a double space or a tab corrupts the mask

`split_mask_field` strips exactly one space character from the head, and only a space:

```python
if head.endswith(' '):
    head = head[:-1]          # exactly one separator space, no more
```

Preserving a *leading* space is the intended behaviour and is right, since the `?s` expansion emits one. But the separator run at the end is not part of the mask, and the previous `' '.join(mask.split())` normalised it away. Measured on this branch against base:

| field | base | this branch |
|---|---|---|
| `'?u?l?l?l?d?d -1 ?u?l?d'` | `['?u?l?l?l?d?d', '-1', '?u?l?d']` | same |
| `'?u?l?l?l?d?d  -1 ?u?l?d'` | `['?u?l?l?l?d?d', '-1', '?u?l?d']` | `['?u?l?l?l?d?d ', '-1', '?u?l?d']` |
| `'?d?d\t-1 ?u?l?d'` | `['?d?d', '-1', '?u?l?d']` | `['?d?d\t', '-1', '?u?l?d']` |

A double space before a charset flag is an ordinary typo, and the result is that every candidate gains a trailing space. It cracks nothing and reports nothing.

```diff
-            head = field[:start]
-            if head.endswith(' '):
-                head = head[:-1]          # exactly one separator space, no more
+            # Strip the whole separator run, not one space: the run is delimiter,
+            # not mask. A LEADING space is still preserved (the ?s expansion emits
+            # one), which is what rejoining tokens would have lost.
+            head = re.sub(r'[ \t]+$', '', field[:start])
```

---

### 3. The agent's sentinel lookup can only ever find a non-sentinel `--`

`install/hashview-agent/hashview-agent.py:488`:

```python
if STATUS_JSON in argv:
    return argv
argv.insert(argv.index('--') if '--' in argv else len(argv), STATUS_JSON)
```

The server emits `--status-json` and `--` as a pair, at `utils.py:1640`, and that is the only place the server emits `--status-json` at all. So whenever a real sentinel is present the function has already returned on the line above. The `argv.index('--')` branch is therefore reachable **only** when there is no sentinel, which means any `--` it finds came from operator data: a mask field of `--`, a mask of `?d?d -- ?d`, or a `-j`/`-k` rule value.

In that case the insert lands in the wrong place. Against an old server with a mask field of `--`, the agent produces `[..., 'hf.txt', '--status-json', '--']`, leaving hashcat with no mask positional at all.

Since the pairing is an invariant the server guarantees, the lookup is not buying anything:

```diff
-    argv.insert(argv.index('--') if '--' in argv else len(argv), STATUS_JSON)
+    # The server emits --status-json paired with the sentinel and nowhere else, so
+    # reaching here means there is no sentinel and any '--' present is operator
+    # data (a mask of '--', a -j/-k rule value). Appending is always correct.
+    argv.append(STATUS_JSON)
```

If you'd rather keep the defensive lookup for a future server that emits a bare sentinel, anchor it to the pairing instead of the bare token, for example by only treating `--` as the sentinel when it is the last element or is immediately preceded by an option.

---

### 4. Smaller items

**The crack shim still leaks values for two flag families.** `tests/e2e/crack/shim/hcshim.py:155` picked up `-1..-4` but not the long forms `--custom-charset1..4`, which the unit tests themselves exercise at `test_mask_argv.py:229`, nor `-d` / `--backend-devices` / `--opencl-device-types`, which an agent configured with `HC_EXTRA_ARGS = '-d 3,4'` would prepend. Their bare values land in positionals and shift the hashfile and wordlist order, which is the exact bug this commit fixed for `--skip` and `--limit`.

**The old-agent claim is only tested in mode 3.** `test_mask_chunk_coverage.py:171` covers the trailing-duplicate case for `-a 3` alone, and mode 7 is in `chunking.MASK_MODES`, so the chunker routinely produces dash-leading sub-masks for it. The description and CHANGELOG both already say hybrid modes need the updated agent, so this is purely a coverage gap rather than an undisclosed one. Parametrising that test over modes 3, 6 and 7 with the mode-3 case asserting success and the others asserting the documented failure would pin the matrix rather than a corner of it.

**Dangling reference.** `hashview/models.py:213` ends with `See #-mask widening.`, which looks like an unfilled issue number.

---

### On the last commit, `75b69df`

The diagnosis is right and I confirmed each premise: the e2e script really does run `pytest -m e2e` with only `tests/security` and `tests/unit` ignored, the marker filters after collection, and `test_mask_chunk_coverage.py` is genuinely the only file in that directory importing the app.

Two reservations:

* **`collect_ignore` is invisible to the gate that exists to catch this.** `hashcat-matrix.yml` greps its own output for `skipped` and fails the job, with a comment saying an always-skipped module would make the gate a no-op that always reports success. A collect-ignored file produces no output at all. If flask ever goes missing from that venv, the flagship test of this PR vanishes and the job goes green. The two conftests this mirrors do not have that problem because nothing greps their output. `HASHCAT_BIN` is set in both matrix invocations, so gating the guard on its absence would keep the stray-invocation benefit while letting a real import failure surface as an error:

```diff
-if importlib.util.find_spec("flask") is None:
+if os.environ.get("HASHCAT_BIN") is None and importlib.util.find_spec("flask") is None:
     collect_ignore = ["test_mask_chunk_coverage.py"]
```

* **It patches a deny-list with another deny-list.** The next directory that imports the app breaks e2e collection again. All `e2e`-marked tests live in `tests/e2e` and `tests/crack` only, so narrowing `tests/run_e2e_compose.sh:139` to those two paths instead of subtracting directories by hand would close the class.

---

### Verified clean

Migration `a4c9e7b21f60` is sound: correct parent, single-headed, `batch_alter_table` is required for SQLite and used correctly, the width guard is round-trip tested, and downgrade truncation is theoretical since `hc_mask` is `String(50)` and MySQL errors rather than truncating under strict mode. Both `DEV_HEAD` pins were bumped here, which #492 forgot. Masks of exactly `-` and `--` behave correctly. Mode 6 and 7 positional paths are server-generated, so there is no wordlist-starts-with-dash risk.

### Cross-PR: migration head collision with #492

Both this revision and `e5d1c7b3a904` declare `down_revision = 'f3b8c1a7d942'`. Whichever merges second needs its parent re-pointed and the head pins bumped again. Already noted in the description; repeating here so it does not get lost at merge time.

bandrel:
Follow-up: I applied the two patches from items 1 and 2 above to a scratch copy of this branch and ran them, so they are not untested suggestions.

**The chunking guard stops the bad case without touching the good one.**

```
options-first mask  '-1 ?u?l?d ?d?d?d'   ->  1 chunk (runs whole)
trailing charset    '?d?d?d -1 ?u?l?d'   ->  1 chunk (runs whole)
bare mask           '?s?d?d?d?d'          ->  330 chunks, unchanged
   dash chunks still produced: -0?d?d?d ... -9?d?d?d
   argv: ['-a', '3', '--status-json', '--', 'h.txt', '-0?d?d?d']
```

So the sentinel behaviour this PR exists to deliver is untouched, and the ten dash chunks still get it.

**The whitespace fix restores the old normalisation and keeps the leading-space case.**

```
'?u?l?l?l?d?d  -1 ?u?l?d'  -> ['?u?l?l?l?d?d', '-1', '?u?l?d']
'?d?d\t-1 ?u?l?d'          -> ['?d?d', '-1', '?u?l?d']
'?u?l?l?l?d?d -1 ?u?l?d'   -> ['?u?l?l?l?d?d', '-1', '?u?l?d']

leading space: mask_attack_argv(3, 'h.txt', None, ' ?d?d', from_chunk=True)
            -> ['-a', '3', 'h.txt', ' ?d?d']
```

The leading space survives and correctly gets no sentinel, which was the reason for slicing by offset in the first place.

**Suite with both patches applied:**

```
tests/unit/test_mask_argv.py tests/unit/test_chunking.py tests/unit/test_crack_shim.py
tests/agent_unit tests/security
280 passed, 19 xfailed
```

No existing test needed editing, which was the acceptance criterion this PR set for itself.

bandrel:
Pushed fixes for the review above, one commit per issue.

| commit | issue |
|---|---|
| `87e8e7a` | chunked custom-charset mask shipping as one literal positional |
| `c9a5058` | double space / tab left inside the mask |
| `0be6924` | agent treating an operator's `--` as the sentinel |
| `f2882bc` | shim leaking long-form charset and device flag values |
| `921386a` | the unfilled issue reference on `chunk_mask` |
| `13f4002` | e2e collecting the whole tree behind a deny-list |
| `9ecf850` | the matrix guard hiding the test it protects |
| `ad4e8bd` | old-agent claim pinned in all three mask modes |
| `b16bf56` | changelog |

Three things worth flagging, because they came out differently from the review.

**The agent `--` bug is reachable on a fully upgraded pair, not just an old server.** I had framed it as an old-server edge. It is not. A mask field of `?d?d -- ?d` splits into three parts whose first does not start with `-`, so no sentinel is emitted and the bare `--` ships as an ordinary positional:

```
['-a', '3', 'hf.txt', '?d?d', '--', '?d']
```

The agent then inserted ahead of that `--`, leaving hashcat with no mask positional at all, which it answers by running its own default mask. Nothing errors.

Since the server emits `--status-json` and `--` as one pair in `mask_attack_argv`, and that is the only place it emits the flag, the early return already covers every real sentinel. So the seek could only ever find operator data, and appending is correct for every argv the server can produce. That commit replaces the two tests that pinned the seek. They built an argv with `--` but no `--status-json`, a shape the server cannot emit, as the neighbouring no-op test already noted.

**The hybrid-mode failure is confirmed by measurement now, not inference.** The parametrized test runs all three mask modes against a real binary. Modes 6 and 7 exit 255 with `No such file or directory`, exactly as the changelog says, and mode 3 still cracks and still reports status.

**The chunking guard does not disarm the fix this branch exists to deliver.** Pinned as a negative control: `?s?d?d?d?d` still plans 330 chunks and the ten dash-leading ones still get the sentinel.

Verification, all on my machine against hashcat v7.1.2:

```
tests/hashcat_matrix          15 passed in 2m42s, no skips
tests/unit + agent_unit + security   2130 passed, 74 xfailed, 0 failures
```

The matrix run matters twice over: it is inside the CI budget, and it produces no `skipped` line, so the workflow's grep gate stays satisfied with the new conftest condition in place.

I also confirmed the e2e narrowing is not a behaviour change: `-m e2e` over the whole tree and `-m e2e tests/e2e` both collect exactly 63 tests.

Issue #499, opened 2026-09-14
Agent sends hashcat stderr as a Python bytes repr, so admin notifications read b'No hashes loaded.\n\n'


Yoshi325:
## Summary

In `run_hashcat` (`install/hashview-agent/hashview-agent.py:477-484`), hashcat's
stderr is captured as `bytes` by `proc.communicate()` and then passed through
`str()` before it is sent to the server. `str(bytes)` is `repr(bytes)`, so the
notification body admins receive is the Python literal, not the message:

```
b'No hashes loaded.\n\n'
```

(`No hashes loaded.` is hashcat's own stderr when a hashfile contains no hash
parseable under the selected `-m` mode. The string exists nowhere in this repo.)

```python
_output, error = proc.communicate()      # error is BYTES
if error:
    LOG.error('Command stderr: %s', error.decode('utf-8', 'replace').strip())  # correct
    if 'hashfile is empty or corrupt' not in str(error):   # matches the repr
        if 'Terminated' in str(error):                     # matches the repr
            sys.exit()
        else:
            api.sendError(str(error))                      # ships the repr
            os.kill(os.getpid(), signal.SIGINT)
```

Line 478 already decodes correctly for the log line, which shows the decode was
known and simply not carried to the other three uses of `error`.

There are two symptoms, one cause:

1. **`api.sendError(str(error))` (line 483)** ships the repr, so the `b`, the
   quotes and the escaped `\n\n` all appear verbatim in the notification.
2. **The comparisons on lines 479-480** substring-match against that same repr.
   Real newlines appear there as the two characters `\n`, and any non-UTF-8 byte
   as `\xNN`. Both suppression checks work today only because the phrases they
   look for happen to be plain ASCII.

## Impact: not Slack-specific

`notify_admins` (`hashview/utils/utils.py:222-248`) hands the identical
`message` to email, Pushover and Slack, so the mangled text lands in all three.

The server cannot repair this. `sendError` ships the body as JSON
(`install/hashview-agent/agent/api/api.py:170`) and `json.dumps` refuses `bytes`
outright, so the `str()` must happen agent-side before transport. By the time
`/v1/error` (`hashview/api/routes.py:1900-1917`) receives it, the body is
already the plain string `"b'No hashes loaded.\n\n'"`. The fix belongs at the
capture site.

Note that agents already deployed in the field will keep emitting the mangled
text until operators re-download and reinstall the agent bundle
(`hashview/agents/routes.py:238`).

## Expected

[...]

PR #500, opened 2026-09-14
:bug: Decode hashcat stderr before reporting it

Yoshi325:
Refs #499.

The agent now reports hashcat's stderr as readable text rather than a Python `bytes` repr.

## The bug

`proc.communicate()` returns `bytes`. `run_hashcat` passed those through `str()`, which is `repr()` for a bytes object, so the notification body admins received was the Python literal:

```
b'No hashes loaded.\n\n'
```

`No hashes loaded.` is hashcat's own stderr when nothing in the hashfile parses under the selected `-m`. The string exists nowhere in this repo.

- **It was never Slack-specific.** `notify_admins()` hands one body to email, Pushover and Slack, so the mangled text landed in all three.
- **The server could not repair it.** `sendError` ships the body as JSON and `json.dumps` refuses `bytes`, so the `str()` has to happen agent-side before transport. By the time `/v1/error` sees it, the body is already the plain string `"b'No hashes loaded.\n\n'"`. The fix belongs at the capture site.

The same `str(error)` appeared in both suppression checks, so those matched against the repr, where a real newline reads as the two characters `\n` and a non-UTF-8 byte as `\xNN`. They worked only because the phrases they look for happen to be plain ASCII.

## The change

One file, `install/hashview-agent/hashview-agent.py`. Decode once into a local and use it for the log line, both comparisons and the report. The `'utf-8', 'replace'` handling is not a new choice: the preceding log line already did exactly that.

The fix also guards the block on a non-empty result. The enclosing `if error:` tests the raw bytes, so a bare newline flush reached the decode and stripped to nothing. Unguarded, that sent admins an alert with an empty body and killed the agent over a blank line.

## Deliberately out of scope

- **No server-side change.** Agents already deployed keep emitting the old text until operators reinstall via the existing agent-bundle download. We did not add a repr-unwrapping hack to `utils.py` to paper over that.
- **The `os.kill(SIGINT)` on a real message is untouched.** That is #281 D, which has its own `strict=True` xfail. It still xfails after this change, and the Testing section shows that rather than assuming it.

## Testing

Six tests in `tests/agent_unit/test_agent_robustness.py`, the file that pins fixed behaviour (the sibling `test_issue_xfail_*` file holds still-broken behaviour).

| Test | Verified failing against the pre-fix source |
|---|---|
| Decoded body, not a repr | yes |
| Multi-line stderr keeps real newlines | yes |
| Non-UTF-8 stderr survives `'replace'` | yes |
| Whitespace-only stderr ignored | yes |
| `hashfile is empty or corrupt` suppressed | guard on the refactor |
| `Terminated` exits quietly | guard on the refactor |

We ran the first four against `HEAD~1`'s source and watched each one fail before restoring the fix, so they catch the bug rather than just describing it. The last two pass before and after by design: they guard the two branches the refactor moves onto a shared local.

```
PYTHONPATH=. pytest tests/agent_unit \
  tests/unit/test_slack_notifications.py \
  tests/unit/test_notifications_routes.py

176 passed, 10 xfailed
```

Zero XPASS matters here. `test_run_hashcat_does_not_selfkill_on_benign_stderr` is `strict=True`, so if this change had quietly absorbed #281 D the suite would fail. It still xfails. Ruff clean on both files.

## Note for the reviewer

Base this against `v0.8.3-dev`, not `main`. GitHub defaults the base to `main`, which is the lagging release branch and does not contain `tests/agent_unit/` at all. The commit says `Refs #499` rather than `Closes` for the same reason: GitHub only honours closing keywords on merge to the default branch.

One follow-on gap went to a comment on #499 instead of into this PR. The phrases this code branches on are themselves unverified against a real binary, because `capture.sh` captures only successful runs and every `tests/fixtures/hashcat/<version>/stderr.txt` is 0 bytes.


i128:
b'thanks'\n\n :P

PR #511, opened 2026-09-15
fix(catalog): offer restore only when the file is missing on disk

i128:
The #383 restore lets an operator replace a wordlist or rule file IN PLACE, keeping the row id and path so every task that references it keeps working. That is the repair for a row which has outlived its file. On the wordlist side it was also offered on rows whose file was fine -- the template said so:

    {# offered on every static row: there is no wordlist edit route, so
       this is also the only way to update one's contents in place #}
    ... title="{{ 'Restore missing file' if gone else 'Replace file' }}"

so overwriting a healthy wordlist was one click, silently changing the candidate set under every task that cites it and under every finished job whose results claim to have used it. Restore is a repair, not an edit.

The upload button now lives inside the `{% if gone %}` branch and the restore dialog only renders for a missing row -- which is how the RULES list has always gated it. Both restore routes also refuse server-side, since they are POST-able directly and the rules route had no such guard either.

Consequence worth stating: there is now NO way to change a static wordlist's contents in place. That capability only ever existed as a side effect of this button, and changing contents means adding a new list (which mints a new row, so tasks must be repointed). Say so if that trade is wrong -- it is the direct cost of the ask, not an oversight.

Three tests added, all three failing against the unfixed code. Two existing tests needed attention:

  * test_wordlists_restore_accepts_a_gzip_upload had become VACUOUS -- it restored a row whose file was present (now refused) and asserted size == 2, which is also the helper's default, so it passed whether or not the upload was applied. It now strands the row first and asserts 3.
  * test_wordlists_restore_invalid_gzip_leaves_the_old_file_intact protected a good file from a bad upload; with restore limited to missing files there is no good file to protect, so it now asserts the equivalent -- a rejected upload does not conjure a broken one, and the row is left missing.

Full suite: 2387 passed, 72 xfailed. ruff clean; bandit exit 0; pylint 9.66.


i128:
blank

Issue #359, opened 2026-07-23
Select multiple hashfiles during job creation, combined into a new hashfile


bandrel:
## Summary
During job creation, the "Use existing" hashfile picker only allows selecting a
single hashfile at a time. Allow selecting **multiple** existing hashfiles and
combine their hashes into one new auto-named hashfile for the job.

## Motivation
Operators frequently accumulate hashes across several uploads (e.g. per-host
pwdumps, staged captures) and want to crack them together in one job. Today the
only workaround is manually merging files outside Hashview and re-uploading,
which loses the per-source hashfiles and duplicates data.

## Current behavior
- UI: single-select radio per row
  `hashview/templates/jobs_assigned_hashfiles.html.j2:238`
  ```html
  <input type="radio" name="hashfile_id" value="{{ hashfile.id }}" ...>
  ```
- Route: reads one `hashfile_id` and assigns it directly to the job
  `hashview/jobs/routes.py:434-438`
  ```python
  elif request.method == 'POST' and request.form.get('hashfile_id'):
      job.hashfile_id = request.form['hashfile_id']
  ```
- A job references exactly one `Hashfiles.id`; there is no notion of combining files.

## Proposed behavior
1. Change the "Use existing" picker from radio to checkboxes so multiple
   hashfiles can be selected.
2. On submit with 2+ selections, create a new `Hashfiles` record and populate it
   by unioning the `HashfileHashes` rows of the selected files, then assign it as
   the job's `hashfile_id`.
3. **Name is auto-generated** (e.g. `combined-<timestamp>` or derived from the
   source names) to avoid collisions — the user does not name it.
4. Single selection keeps today's behavior (assign existing `hashfile_id`
   directly, no new record).

## Constraints
- **Mixed-customer combining is blocked.** The job-creation flow is already
  scoped to a single customer, so the "Use existing" list only shows that
  customer's hashfiles — combining across customers is not reachable in the
  current UI and must remain so. The combined `Hashfiles` record inherits that
  single `customer_id`.

## Deduplication
Hash dedup infrastructure already exists and should be reused:
- `import_hash_only()` dedupes by `(hash_type, MD5(ciphertext))`
  (`hashview/utils/utils.py:544-555`).
- `HashfileHashes` is already many-to-many, so one hash can belong to the
  combined file and its sources.
Combining should produce a distinct set of `hash_id`s (respecting `username`
where present) — no duplicate `HashfileHashes` rows in the combined file.

## No schema changes
[...]

Issue #298, opened 2026-06-30
Security (MEDIUM): missing CSRF protection on form-reading routes


bandrel:
## Summary

No global CSRF protection is installed, so state-changing routes that read `request.form` (or act on GET) directly — instead of going through `form.validate_on_submit()` — accept forged cross-site requests.

## Details

- `hashview/__init__.py` only registers a `csrf_token()` template global; there is **no** `CSRFProtect(app)`.
- Routes that bypass `validate_on_submit()` and read `request.form` directly (or mutate on GET) are unprotected, including:
  - `/customers/edit`
  - `/customers/delete`
  - `/hashfiles/delete`
  - the GET-based agent authorize / deauthorize / delete routes
- Routes that **do** use `form.validate_on_submit()` (e.g. `/customers/add`) are protected, since Flask-WTF validates the CSRF token there.

## Severity

MEDIUM — CSRF on authenticated state-changing actions (delete customer/hashfile, authorize/delete agents).

## Reproduction / proof

Demonstrated in `tests/security/test_security_hardening.py`:
- `test_csrf_protected_create_rejects_without_token_and_accepts_with` (positive control on `/customers/add`)
- `test_csrf_unprotected_edit_route_is_a_finding` — strict xfail proving `/customers/edit` mutates state with no token.

The finding test is `xfail(strict=True)`; fixing this flips it to a pass and the marker should be removed.

## Suggested remediation

- Install `flask_wtf.CSRFProtect(app)` globally and ensure every state-changing route either uses a CSRF-validated form or an explicit token check; convert GET-based mutations (agent authorize/delete) to POST with tokens.

Found during test-architecture hardening (PR #296).

bandrel:
Regression test added in **#300** (`tests/security/test_csrf_xfail.py`): a strict xfail proving a token-less POST to `/customers/edit` mutates state (no global `CSRFProtect`). When a CSRF gate is added, the strict xfail flips to a visible pass and the `xfail` marker should be removed.

Issue #206, opened 2026-06-13
IDOR: a job can be assigned a hashfile belonging to another customer


bandrel:
**Severity:** High — broken customer data segregation (IDOR).

**Location:** \`hashview/jobs/routes.py:400-404\` (in \`jobs_assigned_hashfile\`)

```python
elif request.method == 'POST' and request.form.get('hashfile_id'):
    job.hashfile_id = request.form['hashfile_id']   # no ownership/customer check
    db.session.commit()
    return redirect("/jobs/" + str(job.id)+"/notifications")
```

The posted \`hashfile_id\` is assigned to the job with no verification that the hashfile belongs to \`job.customer_id\`. Any authenticated user can attach **another customer's** hashfile to their job by posting that id.

**Confirmed behavior:** With job J in customer A and hashfile H in customer B, \`POST /jobs/<J>/assigned_hashfile/\` with \`hashfile_id=<H>\` sets \`job.hashfile_id = H\`. Verified with a Flask test-client reproduction on \`v0.8.3-dev\` (\`job.customer=1, hf.customer=2, job.hashfile_id=2\`).

**Fix:** validate the selected hashfile belongs to the job's customer before assigning, e.g.:
```python
hf = Hashfiles.query.filter_by(id=request.form['hashfile_id'], customer_id=job.customer_id).first()
if hf is None:
    flash('Invalid hashfile selection.', 'danger')
    return redirect(url_for('jobs.jobs_assigned_hashfile', job_id=job.id))
job.hashfile_id = hf.id
```

i128:
Im not sure this is a security issue.

PR #521, opened 2026-09-16
feat(jobs): select multiple existing hashfiles when building a job (#359)

bandrel:
## Summary

Makes the "Use existing" hashfile picker on `/jobs/<id>/assigned_hashfile/` multi-select. Selecting two or more hashfiles combines them into one new auto-named hashfile (`combined-YYYYMMDD-HHMMSS`) which is assigned to the job; selecting exactly one keeps today's behaviour and creates no new record. The combined file is a normal saved `Hashfiles` record, so it is reusable and is removed by the usual data-retention sweep. No schema change, no migration, no `/v1` API change.

Closes #359

Refusals surface as flashed errors and change nothing:

- sources that disagree on hash type, or whose representative hash is unresolvable — a job derives its hash type from the first linked hash (`hashfile_hash_type` → `build_hashcat_command`), so a mixed-type file would silently crack only one type
- sources that disagree on `hex_salt`, a single per-file flag a combined file cannot honour both ways
- a source with no hashes, a missing id, or an id belonging to another customer

Rows are deduplicated on `(hash_id, username)`, so the same hash under two usernames is kept twice — a real distinction for pwdump/NTDS sources — and first-seen order is preserved across sources.

### Three things I'd like reviewer eyes on

**1. Changing radio → checkbox changes what an empty selection posts.** A radio group always submits its checked value; clearing every checkbox submits nothing at all, and `hf_form_existing` has no CSRF token and its submit button lives outside the form. So the form now carries a hidden `hf_source` marker to distinguish "existing form submitted with nothing selected" from an unrelated POST. The old `hashfile_id`-based condition is retained as an alternative so a bare `hashfile_id` post still works.

**2. Nothing is pre-selected any more.** The template used to pre-check the job's current hashfile (or the first row for a new job). That was invisible with a radio, because picking any row cleared the others — but with checkboxes it means clicking the hashfile you actually want silently combines it with the pre-checked one. The current hashfile is now shown as a non-interactive `current` badge instead. This is a small UX change on the edit path and I'm happy to revisit the presentation.

**3. Scaling.** Sources are streamed with `yield_per` and inserted with `bulk_insert_mappings` per `_IMPORT_CHUNK_SIZE`, mirroring `import_hashfilehashes`, so combining two NTDS-scale dumps doesn't build millions of in-flight ORM objects or one multi-million-row `INSERT`. It is deliberately **one transaction** (flush per chunk, single commit) so a mid-combine failure can't leave a partially filled hashfile behind — there's a comment saying so, since the tempting "improvement" is to add per-chunk commits. Related: #364 (moving hashfile import to the background) would apply to this path too.

### Security notes

- The existing-hashfile branch now validates the selected id against `job.customer_id`. That path previously assigned `request.form['hashfile_id']` unvalidated. This **narrows #206 on this route but does not close it** — `Jobs.query.get(job_id)` is still unscoped.
- This does **not** add CSRF to `hf_form_existing` (#298 is repo-wide and out of scope here), but it's worth stating plainly that it widens that gap's blast radius on this form: a forged POST could previously only flip `job.hashfile_id`; it can now also create a `Hashfiles` row plus a large number of `HashfileHashes` rows. Happy to add a `FlaskForm` here if you'd rather not ship that asymmetry.

### Known limitation

Only the *representative* (first) link's hash type is validated, consistent with `hashfile_hash_type` / `build_hashcat_command`. A dangling link at position 2+ of a source is copied into the combined file as-is. Silently skipping unresolvable links seemed worse than a documented limit, but say the word if you'd prefer the filter.

## Checklist

- [x] This PR targets `v0.8.3-dev` (the current dev branch), **not** `main`
- [x] Commit subjects follow `type(scope): subject` (see [CONTRIBUTING.md](../CONTRIBUTING.md#commit-messages))
- [x] `scripts/preflight.sh` passes locally
- [x] Pushed through the pre-push hook (not `--no-verify`)
- [x] New/changed behavior has test coverage
- [ ] If a route changed: `hashview/api_docs/openapi.yaml` is updated
- [ ] If a migration was added: it's MySQL/MariaDB-safe, and `DEV_HEAD` in
      `tests/run_migration_e2e.sh` is bumped
- [x] If user-visible: `CHANGELOG.md` has an entry under `## Current Release`
- [x] N/A items above are just left unchecked, not deleted

`openapi.yaml` is unchanged because this adds no `/v1` route — `jobs_assigned_hashfile` is a web-UI route only.

## Testing

`scripts/preflight.sh` → ruff, bandit (vs baseline), OpenAPI spec validation, pylint, and `pytest tests/unit` all pass. Full gate `pytest tests/unit tests/security tests/agent_unit` → **2428 passed, 2 skipped, 76 xfailed**.

Coverage added: 22 tests for the helper (`tests/unit/test_combine_hashfiles.py`) and 9 for the route/template (`tests/unit/test_jobs_assigned_hashfile.py`), including the dedup-across-chunk-boundary case, first-seen ordering under a patched chunk size, a forced `SQLAlchemyError` on commit, the empty-selection and other-customer refusals, and a three-hashfile render test asserting nothing comes back pre-checked.

One unrelated thing I hit while working on this: `tests/unit/test_stranded_jobtask_reclaim.py::test_a_live_agents_row_is_left_alone` fails on any machine whose local time isn't UTC. It seeds `last_checkin` with `datetime.now()` while the sweep compares against `utcnow()`, so off-UTC the agent looks stale and gets reclaimed. It passes under `TZ=UTC`, which is why CI is green. Unrelated to this PR — happy to file it separately.


bandrel:
Follow-up commit 3ac92ff: the source reads now use keyset paging on `HashfileHashes.id` instead of `yield_per`.

The `yield_per` was there to bound memory when combining NTDS-scale hashfiles, but it can't do that here — the `mysql-connector-python` dialect reports `supports_server_side_cursors = False`, so the driver buffers the whole result set client-side regardless. That's the same finding already recorded in `settings/routes.py:73-80`, where a hash export measured a 114 MB peak with `yield_per` and went flat at ~4 MB after switching to primary-key paging. This mirrors that path and reuses `_IMPORT_CHUNK_SIZE` as the page size.

Per-chunk insert and single-commit semantics are unchanged, as are the first-seen ordering and cross-chunk dedup guarantees — both still covered by their existing tests, unmodified.

Full gate: 2429 passed, 2 skipped, 76 xfailed.

